PR #4, opened 2022-12-02
Fix some issue to make batch inference easy for predictor and decoder.

TeaPoly:
I noticed that the inference process uses single sample. When I try to using batch (dynamic or static) inference. The Encoder of the Paraformer model can easy to use the batch inference, but the predictor and decoder will have some problems. So I fix some minor issue, to meet the needs of batch processing in the future


langgz (approved):
Thank you for your contributions. I have tested your changes, it works. We will merge them.

PR #5, opened 2022-12-02
update modelscope details

Lizerui9926:


langgz (approved):
modelscope recipe update

PR #44, opened 2023-01-26
better asr text result with json style output

fangd123:


langgz (approved):
Thank for your advice.

PR #45, opened 2023-01-29
add 8k paraformer recipe

hnluo:


hnluo:
add 8k paraformer recipe

langgz (approved):

PR #147, opened 2023-02-22
Add sentence timestamp support

dingbig:
Added support for sentence timestamp, which is particularly useful for applications such as lyrics and subtitles.
I added a function named time_stamp_sentence in timestamp_tool.py,


R1ckShi (approved):
supports return items in setences seperated by punc with merged timestamps

PR #267, opened 2023-03-20
fix bug for onnx paraformer-long

R1ckShi:
fix bug for onnx paraformer-long in timestamp inference for code-switch transcription


langgz (commented):
libtorch同步修改

langgz (approved):

PR #290, opened 2023-03-24
Read audio_data to buf when speaking is False for non-stream inferring

veelion:
For non-stream inferring, this can reduce sending `speaking=True` status to server, only send audio data with `speaking=False` status.


dyyzhmm:
Hi @veelion , thanks for your attention.

In my opinion, this PR is not a bug, but an another method differs from [original design](https://github.com/alibaba-damo-academy/FunASR/blob/main/funasr/runtime/python/grpc/proto/workflow.png).

I thinks this changes will leads to a **latency** from network when sending many chunks of audio data in one request.
Do you agree?   

Waiting for your response.

veelion:
Hi @dyyzhmm ,  I agree with with you very much, a network latency occurs while sending large data in one request. 

We can let users decide whether to accept this delay when he want to send large audio, if he can't accpet this, just split it to many chunks and send them according the original design.

This PR does not affect the original design, but give user a more choice. what do you think?

langgz:
> Hi @dyyzhmm , I agree with with you very much, a network latency occurs while sending large data in one request.
> 
> We can let users decide whether to accept this delay when he want to send large audio, if he can't accpet this, just split it to many chunks and send them according the original design.
> 
> This PR does not affect the original design, but give user a more choice. what do you think?

How to set whether to send large audio, is there a flag?

dyyzhmm:
@veelion  Good idea.    Just as @langgz mentioned above, can you @veelion  make a flag for user's choice ( in  .cc & readme )  and make a new PR?

veelion:
For client, this is a example of the way to send whole audio to server in one request:
```
async def send(channel, data, speaking, isEnd):
    stub = RecognizeStub(channel)
    req = paraformer_pb2.Request()
    if data:
        req.audio_data = data
    req.user = 'zz'
    req.language = 'zh-CN'
    req.speaking = speaking
    req.isEnd = isEnd
    q = queue.SimpleQueue()
    q.put(req)
    return stub.Recognize(iter(q.get, None))

# send the audio data once
with grpc.insecure_channel(grpc_uri) as channel:
        response = await send(channel, data, speaking=False, isEnd=False)
        resp = response.next()
        text = ''
        if 'decoding' == resp.action:
            resp = response.next()
            if 'finish' == resp.action:
                text = json.loads(resp.sentence)['text']
        response = await send(channel, None, False, True) 
print(text)
```

I will add this example to readme.

lyblsgo (commented):
Read audio_data to buf when speaking is False for non-stream inferring

lyblsgo (approved):

PR #353, opened 2023-04-14
support arbitrary sampling rate

lyblsgo:


langgz (approved):

funasr/runtime/onnxruntime/src/Audio.cpp
```diff
@@ -3,11 +3,96 @@
...
+#include <assert.h>
 
 #include "Audio.h"
+#include "precomp.h"
 
 using namespace std;
 
+// see http://soundfile.sapp.org/doc/WaveFormat/
+// Note: We assume little endian here
+struct WaveHeader {
```
csukuangfj, line 16: I suggest that you mention in some way the code is copied from
https://github.com/k2-fsa/sherpa-onnx/blob/master/sherpa-onnx/csrc/wave-reader.cc#L20
for `struct WaveHeader`.

PR #830, opened 2023-08-10
fix bug for timestamp inference

R1ckShi:


R1ckShi:
update timestamp inference

PR #1532, opened 2024-03-22
Expose the max_end_silence_time to the user

JavanTang:
Using the VAD model, it is common about the change 'max_end_silence_time'.


LauraGPT:
This is a demo: https://github.com/alibaba-damo-academy/FunASR/blob/main/examples/industrial_data_pretraining/paraformer/demo.py#L12

JavanTang:
@LauraGPT Thanks for the reply. Is this only used during the AutoModel structure?  If I want to change the 'max_end_silence_time', do I need to reload the overall model?

LauraGPT:
> @LauraGPT Thanks for the reply. Is this only used during the AutoModel structure? If I want to change the 'max_end_silence_time', do I need to reload the overall model?

OK

PR #1791, opened 2024-06-07
Accelerate cif

dtlzhuangz:
The pr accelerates cif and cif_export. The for loop in the cif bottlenecks the inference time of onnx model. 

torch profiler of cif:
image
torch profiler of cif_v1:
image
Time consumption of onnxruntime-gpu
image

The precision:
image

### DIFF
--- a/funasr/models/paraformer/cif_predictor.py
+++ b/funasr/models/paraformer/cif_predictor.py
@@ -661,14 +661,13 @@ def cif(hidden, alphas, threshold):
     return torch.stack(list_ls, 0), fires
 
 
-def cif_v1(hidden, alphas, threshold):
+def cif_wo_hidden_v1(alphas, threshold, return_fire_idxs=False):
+    batch_size, len_time = alphas.size()
+    device = alphas.device
+    dtype = alphas.dtype
 
-    device = hidden.device
-    dtype = hidden.dtype
-    batch_size, len_time, hidden_size = hidden.size()
     threshold = torch.tensor([threshold], dtype=alphas.dtype).to(alphas.device)
 
-    frames = torch.zeros(batch_size, len_time, hidden_size, dtype=dtype, device=device)
     fires = torch.zeros(batch_size, len_time, dtype=dtype, device=device)
 
     prefix_sum = torch.cumsum(alphas, dim=1)
@@ -682,7 +681,19 @@ def cif_v1(hidden, alphas, threshold):
     fire_idxs = dislocation_diff > 0
     fires[fire_idxs] = 1
     fires = fires + prefix_sum - prefix_sum_floor
+    if return_fire_idxs:
+        return fires, fire_idxs
+    return fires
 
+
+def cif_v1(hidden, alphas, threshold):
+    fires, fire_idxs = cif_wo_hidden_v1(alphas, threshold, return_fire_idxs=True)
+
+    device = hidden.device
+    dtype = hidden.dtype
+    batch_size, len_time, hidden_size = hidden.size()
+    frames = torch.zeros(batch_size, len_time, hidden_size,
+                         dtype=dtype, device=device)
     prefix_sum_hidden = torch.cumsum(
         alphas.unsqueeze(-1).tile((1, 1, hidden_size)) * hidden, dim=1
     )
@@ -698,7 +709,8 @@ def cif_v1(hidden, alphas, threshold):
 
     remains = fires - torch.floor(fires)
     remain_frames = (
-        remains[fire_idxs].unsqueeze(-1).tile((1, hidden_size)) * hidden[fire_idxs]
+        remains[fire_idxs].unsqueeze(-1).tile((1,
+                                               hidden_size)) * hidden[fire_idxs]
     )
 
     shift_remain_frames = torch.roll(remain_frames, 1, dims=0)
--- a/funasr/models/paraformer/cif_predictor.py
+++ b/funasr/models/paraformer/cif_predictor.py
@@ -80,7 +80,7 @@ class CifPredictor(torch.nn.Module):
                     hidden, alphas, token_num, mask=mask
                 )
 
-            acoustic_embeds, cif_peak = cif(hidden, alphas, self.threshold)
+            acoustic_embeds, cif_peak = cif_v1(hidden, alphas, self.threshold)
 
             if target_length is None and self.tail_threshold > 0.0:
                 token_num_int = torch.max(token_num).type(torch.int32).item()
@@ -245,7 +245,7 @@ class CifPredictorV2(torch.nn.Module):
                         hidden, alphas, token_num, mask=None
                     )
 
-            acoustic_embeds, cif_peak = cif(hidden, alphas, self.threshold)
+            acoustic_embeds, cif_peak = cif_v1(hidden, alphas, self.threshold)
             if target_length is None and self.tail_threshold > 0.0:
                 token_num_int = torch.max(token_num).type(torch.int32).item()
                 acoustic_embeds = acoustic_embeds[:, :token_num_int, :]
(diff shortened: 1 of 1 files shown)


TeaPoly:
I think whether this part of the optimization function can be extended to the `cif_wo_hidden` function, because the monotonic forced alignment model will use this part of the code.

```python

import torch


def cif_wo_hidden_v1(alphas, threshold, return_fire_idxs=False):
    batch_size, len_time = alphas.size()
    device = alphas.device
    dtype = alphas.dtype

    threshold = torch.tensor([threshold], dtype=alphas.dtype).to(alphas.device)

    fires = torch.zeros(batch_size, len_time, dtype=dtype, device=device)

    prefix_sum = torch.cumsum(alphas, dim=1)
    prefix_sum_floor = torch.floor(prefix_sum)
    dislocation_prefix_sum = torch.roll(prefix_sum, 1, dims=1)
    dislocation_prefix_sum_floor = torch.floor(dislocation_prefix_sum)

    dislocation_prefix_sum_floor[:, 0] = 0
    dislocation_diff = prefix_sum_floor - dislocation_prefix_sum_floor

    fire_idxs = dislocation_diff > 0
    fires[fire_idxs] = 1
    fires = fires + prefix_sum - prefix_sum_floor
    if return_fire_idxs:
        return fires, fire_idxs
    return fires


def cif_v1(hidden, alphas, threshold):
    fires, fire_idxs = cif_wo_hidden_v1(alphas, threshold, return_fire_idxs=True)

    device = hidden.device
    dtype = hidden.dtype
    batch_size, len_time, hidden_size = hidden.size()
    frames = torch.zeros(batch_size, len_time, hidden_size,
                         dtype=dtype, device=device)
    prefix_sum_hidden = torch.cumsum(
        alphas.unsqueeze(-1).tile((1, 1, hidden_size)) * hidden, dim=1
    )

    frames = prefix_sum_hidden[fire_idxs]
    shift_frames = torch.roll(frames, 1, dims=0)

    batch_len = fire_idxs.sum(1)
    batch_idxs = torch.cumsum(batch_len, dim=0)
    shift_batch_idxs = torch.roll(batch_idxs, 1, dims=0)
    shift_batch_idxs[0] = 0
    shift_frames[shift_batch_idxs] = 0

    remains = fires - torch.floor(fires)
    remain_frames = (
        remains[fire_idxs].unsqueeze(-1).tile((1,
                                               hidden_size)) * hidden[fire_idxs]
    )

    shift_remain_frames = torch.roll(remain_frames, 1, dims=0)
    shift_remain_frames[shift_batch_idxs] = 0

    frames = frames - shift_frames + shift_remain_frames - remain_frames

    max_label_len = batch_len.max()

    frame_fires = torch.zeros(
        batch_size, max_label_len, hidden_size, dtype=dtype, device=device
    )
    indices = torch.arange(max_label_len, device=device).expand(batch_size, -1)
    frame_fires_idxs = indices < batch_len.unsqueeze(1)
    frame_fires[frame_fires_idxs] = frames
    return frame_fires, fires

```

TeaPoly:
May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.

dtlzhuangz:
> May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.

I think it is because there is a little difference between cif_export and cif. Actually, I write the code follow the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor althought it does not use floor explicitly. Does the difference matter in training?

LauraGPT (approved):

LauraGPT:
> > May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.
> 
> I think it is because there is a little difference between cif_export and cif. Actually, I write the code following the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor. Does the difference matter in training?

@R1ckShi Please check it?

R1ckShi:
no matter which accumulate function you use, in the training stage of paraformer we use the real number of label to re-scale the sum of alphas, for ensuring the attention loss calculation has the same logits and label. so maybe you miss the re-scale function in cif training forward

R1ckShi:
> > May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.
> 
> I think it is because there is a little difference between cif_export and cif. Actually, I write the code following the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor. Does the difference matter in training?

no matter which accumulate function you use, in the training stage of paraformer we use the real number of label to re-scale the sum of alphas, for ensuring the attention loss calculation has the same logits and label. so using it in training stage requires a re-scale process.

dtlzhuangz:
> > > May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.
> > 
> > 
> > I think it is because there is a little difference between cif_export and cif. Actually, I write the code following the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor. Does the difference matter in training?
> 
> no matter which accumulate function you use, in the training stage of paraformer we use the real number of label to re-scale the sum of alphas, for ensuring the attention loss calculation has the same logits and label. so using it in training stage requires a re-scale process.

OK, I will modify the cif.

R1ckShi:
> > > > May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.
> > > 
> > > 
> > > I think it is because there is a little difference between cif_export and cif. Actually, I write the code following the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor. Does the difference matter in training?
> > 
> > 
> > no matter which accumulate function you use, in the training stage of paraformer we use the real number of label to re-scale the sum of alphas, for ensuring the attention loss calculation has the same logits and label. so using it in training stage requires a re-scale process.
> 
> OK, I will modify the cif.

sorry i made a mistake, rescale is conducted outside of cif(). so maybe you can check whether your code works correctly here @TeaPoly 
funasr/models/paraformer/cif_predictor.py: 249
```python
            acoustic_embeds, cif_peak = cif_v1(hidden, alphas, self.threshold)
            if target_length is None and self.tail_threshold > 0.0:
                token_num_int = torch.max(token_num).type(torch.int32).item()
                acoustic_embeds = acoustic_embeds[:, :token_num_int, :]
```

dtlzhuangz:
> > > > > May I ask if this part of the code has tried to train the model? When I use this part of the code for training, sometimes the token length and the time dimension of the `acoustic_embeds` will have a difference of one.
> > > > 
> > > > 
> > > > I think it is because there is a little difference between cif_export and cif. Actually, I write the code following the cif_export for onnx inference. The difference is that cif uses torch.round for len while cif_export is floor. Does the difference matter in training?
> > > 
> > > 
> > > no matter which accumulate function you use, in the training stage of paraformer we use the real number of label to re-scale the sum of alphas, for ensuring the attention loss calculation has the same logits and label. so using it in training stage requires a re-scale process.
> > 
> > 
> > OK, I will modify the cif.
> 
> sorry i made a mistake, rescale is conducted outside of cif(). so maybe you can check whether your code works correctly here @TeaPoly funasr/models/paraformer/cif_predictor.py: 236
> 
> ```python
>             if target_length is not None:
>                 alphas *= (target_length / token_num)[:, None].repeat(1, alphas.size(1))
>             elif self.tail_threshold > 0.0:
>                 if self.tail_mask:
>                     hidden, alphas, token_num = self.tail_process_fn(
>                         hidden, alphas, token_num, mask=mask
>                     )
>                 else:
>                     hidden, alphas, token_num = self.tail_process_fn(
>                         hidden, alphas, token_num, mask=None
>                     )
> 
>             acoustic_embeds, cif_peak = cif_v1(hidden, alphas, self.threshold)
>             if target_length is None and self.tail_threshold > 0.0:
>                 token_num_int = torch.max(token_num).type(torch.int32).item()
>                 acoustic_embeds = acoustic_embeds[:, :token_num_int, :]
> ```

Error occurs when I finetune the model using this code. And I have modified the code to make training run. Let me train a model to verify the correctness of training result. Sorry for the trouble.

PR #1811, opened 2024-06-12
Update cif_predictor.py

dtlzhuangz:
Modify the max_label_len to match the length in training. Please check whether the training process and result is correct. If there is something wrong, I suggest back the cif_v1 to cif in https://github.com/dtlzhuangz/FunASR/blob/main/funasr/models/paraformer/cif_predictor.py#L248.


dtlzhuangz:
Under extreme circumstances, torch.cumsum have precision degradation causing the label_len misalign with token_num. We use float64 and alphas.sum() instead. We tested under two datasets. Base WER: 13.87， 2.91； new WER: 13.4, 2.99.

LauraGPT (approved):

dtlzhuangz:
@LauraGPT @R1ckShi 这个pr里的代码原本修复了training和inference的bug但是在https://github.com/modelscope/FunASR/commit/45d7aa9004763684fb748ee17942ecba81042201 应该是被不小心丢掉了

PR #2329, opened 2024-12-23
utils.install_model_requirements: support installing with uv

zmwangx:
[uv][1] is a new high performance pip replacement that's been quite popular recently.

I was able to install and use FunASR with uv, except one problem: `utils.install_model_requirements` executes `pip install` directly, which causes a `FileNotFoundError` when using uv. This PR runs [uv's pip-compatible interface][2] when `pip` is not found on `PATH` but `uv` is.

[1]: https://docs.astral.sh/uv/
[2]: https://docs.astral.sh/uv/pip/packages/

### DIFF
--- a/funasr/utils/install_model_requirements.py
+++ b/funasr/utils/install_model_requirements.py
@@ -1,15 +1,10 @@
+import shutil
 import subprocess
 
 
 def install_requirements(requirements_path):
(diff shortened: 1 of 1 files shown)


LauraGPT (approved):
Thanks

PR #2373, opened 2025-02-01
fix: resolve unexpected 'out of memory' issue in multi-GPU setup

BienBoy:
Fixed a bug where calling `torch.cuda.empty_cache()` caused extra memory usage on 'cuda:0', leading to unexpected 'out of memory' errors in multi-GPU environments.

To reproduce the issue: 
1. On a multi-GPU machine, run a model on `cuda:0` until its memory usage reaches 100%.
2. Run `paraformer-zh` on another GPU, causing the out-of-memory error.

The issue may be the bug described in [#1145](https://github.com/modelscope/FunASR/issues/1145).

Reference:
- https://github.com/pytorch/pytorch/issues/25752
- https://github.com/pytorch/pytorch/issues/144025

### DIFF
--- a/funasr/auto/auto_model.py
+++ b/funasr/auto/auto_model.py
@@ -366,7 +366,8 @@ class AutoModel:
         if pbar:
             # pbar.update(1)
             pbar.set_description(f"rtf_avg: {time_escape_total/time_speech_total:0.3f}")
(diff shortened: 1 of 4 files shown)


LauraGPT (approved):
Thanks

PR #2454, opened 2025-03-26
Update model.py to fix "IndexError: index 1 is out of bounds for dimension 1 with size 0"

Isuxiz:
Avoid exception of "IndexError: index 1 is out of bounds for dimension 1 with size 0".

This is caused when sensevoice model output result contains no text, like below:
```
res=[{'key': 'rand_key_1t9EwL56nGisi', 'text': '<|zh|><|NEUTRAL|><|Speech|><|woitn|>'}]
```

Just skip segments like this can fix this bug.

### DIFF
--- a/funasr/models/sense_voice/model.py
+++ b/funasr/models/sense_voice/model.py
@@ -925,6 +925,11 @@ class SenseVoiceSmall(nn.Module):
                     if tok_ls: token_ids.extend(tok_ls)
                     else: token_ids.append(124)
 
+                if len(token_ids) == 0:
+                    result_i = {"key": key[i], "text": text}
+                    results.append(result_i)
+                    continue
+
                 logits_speech = self.ctc.softmax(encoder_out)[i, 4 : encoder_out_lens[i].item(), :]
                 pred = logits_speech.argmax(-1).cpu()
                 logits_speech[pred == self.blank_id, self.blank_id] = 0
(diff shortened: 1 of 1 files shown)


passerbya:
Traceback (most recent call last):
  File "/root/miniconda3/envs/sensevoice/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/root/miniconda3/envs/sensevoice/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/data2/workspace/egs_vocal_extractor/data/speech_det.py", line 205, in process_audio_task
    res = model.generate(
  File "/root/miniconda3/envs/sensevoice/lib/python3.10/site-packages/funasr/auto/auto_model.py", line 306, in generate
    return self.inference_with_vad(input, input_len=input_len, **cfg)
  File "/root/miniconda3/envs/sensevoice/lib/python3.10/site-packages/funasr/auto/auto_model.py", line 516, in inference_with_vad
    t[0] += vadsegments[j][0]
TypeError: can only concatenate str (not "int") to str
你这个修复会报这个错误

passerbya:
这三个地方应该是timestamp_new.append([start, end])

Isuxiz:
> 图片
> 
> 这三个地方应该是timestamp_new.append([start, end])

Yes, there is a problem here. In my project, I try to align the timestamp with the token directly to avoid re-splitting the text by token later. I shouldn't submit this commit to this branch for PR, and I will undo this submission later.

LauraGPT (approved):

PR #2556, opened 2025-06-19
Update server.py

WThirteen:
Fix the error that occurs when there are no text elements in the audio and an exception is thrown.


WThirteen:
当语音识别文件中，并没有文字时，我们期望得到一个
image
而并非是如下的报错
image

LauraGPT (approved):

PR #2957, opened 2026-06-01
fix(cpu): enable CPU support by bypassing vLLM in realtime server

Manamama-Gemini-Cloud-AI-01:
This PR adds CPU support to the FunASR Nano real-time server by automatically bypassing vLLM when the device is set to 'cpu'. It also includes necessary interface adapters to handle the output differences between standard AutoModel (PyTorch) and AutoModelVLLM.


Manamama-Gemini-Cloud-AI-01:
We shall try to create a better PR with this then.

PR #2958, opened 2026-06-01
fix(cpu): enable robust CPU support and address PR feedback

Manamama-Gemini-Cloud-AI-01:
This PR enables robust CPU support for the FunASR Nano real-time server. 

Key changes:
- Dynamically bypasses vLLM when the device is set to a non-CUDA device (e.g., 'cpu').
- Implements robust inference result extraction to handle different return types between `AutoModel` (PyTorch) and `AutoModelVLLM`.
- Improves tokenizer access safety and input buffer handling (AI review feedback).
- Optimizes hallucination detection.


LauraGPT (commented):
Thanks for the contribution! A few concerns:

1. **The example files (client_python.py, demo_vllm.py, serve_realtime_ws.py) appear to be full replacements** (+1015 additions, 0 deletions). This means the diff shows the entire file as new, making it very hard to review what actually changed. Could you rebase on the latest main so the diff shows only your actual modifications?

2. **The encoder_conf None check** in `fsmn_vad_streaming/model.py` is a good fix — this part looks clean.

3. **Scope**: The PR title says "CPU support" but it also includes hallucination detection optimization, tokenizer changes, and buffer handling. Could you split these into separate PRs? It makes reviewing much easier.

Please rebase on latest main so we can see the incremental changes. The CPU fallback (bypassing vLLM on non-CUDA) is a useful feature worth merging, but we need to review it as a clean diff.

LauraGPT:
Thanks for working on CPU support for the Fun-ASR-Nano realtime server — this is genuinely useful for users without a GPU. 🙏

A few things to help us review and land it:

1. It's a large diff (+1000 lines). Could you briefly outline the main pieces — the vLLM-bypass path for non-CUDA devices, the realtime WS server changes, and the `fsmn_vad_streaming/model.py` change — and what motivated each?
2. How did you test it end-to-end on CPU (which model, sample audio, expected vs. actual output)? A short repro would let us verify on our side.
3. The `fsmn_vad_streaming/model.py` change touches a path shared by GPU users — could you confirm it's behavior-preserving on CUDA?

Once that's clear we can test it on a CPU box and move it forward. Thanks again!

LauraGPT:
Thanks for tackling CPU support for the Fun-ASR-Nano realtime server 🙏 A status update:

- The FSMN-VAD `encoder_conf is None` guard from this PR is **already in `main`** (shipped in v1.3.10).
- CPU / edge inference is now well-served by the new **llama.cpp / GGUF runtime** — a single self-contained binary, zero Python, built-in VAD (like whisper.cpp): https://github.com/modelscope/FunASR/tree/main/runtime/llama.cpp . That covers the no-GPU use case more cleanly than maintaining Python serving examples.
- This PR also currently has merge conflicts and hasn't been updated in ~3 weeks.

Closing for now given the above — but thank you for the contribution! If there's a specific CPU gap the llama.cpp runtime doesn't cover, please feel free to open a focused, rebased PR. 🙏

Issue #2948, opened 2026-05-28
vllm离线服务推理，长音频报错


jun-zhang:
Notice: In order to resolve issues efficiently, please follow the template and include reproducible details.
（注意：为了更加高效率解决您遇到的问题，请按照模板提问，补充可复现细节。）

## 🐛 Bug

vllm离线服务推理，长音频报错

## To Reproduce

Steps to reproduce the behavior. Always include the exact command you ran.

CUDA_VISIBLE_DEVICES=1 python serve_vllm.py --model FunAudioLLM/Fun-ASR-Nano-2512 --gpu-memory-utilization 0.5 --port 8899

## Code sample

```python

```

## Error logs

EngineCore pid=3370471) WARNING 05-28 09:25:52 [jit_monitor.py:103] Triton kernel JIT compilation during inference: _compute_slot_mapping_kernel. This causes a latency spike; consider extending warmup to cover this shape/config.
Rendering prompts:  78%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████                                         | 642/818 [00:00<00:00, 2877.30it/s]/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [19,0,0], thread: [9,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [19,0,0], thread: [10,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [6,0,0], thread: [37,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [13,0,0], thread: [109,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [13,0,0], thread: [110,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
[...]

PR #2974, opened 2026-06-11
fix: avoid CUDA crash from repetition_penalty in Fun-ASR-Nano vLLM prompt-embeds mode (#2948)

SuperMarioYL:
## Summary

The Fun-ASR-Nano vLLM serving paths run the engine with `enable_prompt_embeds=True`, so each request carries precomputed audio/text **embeddings** and has **no prompt token IDs**. vLLM applies `repetition_penalty` by scattering over the prompt's token IDs, so any value other than `1.0` indexes an empty token-id tensor and aborts the engine with:

```
ScatterGatherKernel.cu:163: operator(): Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
```

This is the crash reported in #2948. The batch `generate()` entry already defaulted to `repetition_penalty=1.0`, but three other prompt-embeds paths still forwarded `1.3` and crashed:

- `funasr/bin/_server_app.py` — the OpenAI-compatible / REST server passes `repetition_penalty=1.3` straight into the vLLM engine.
- `funasr/models/fun_asr_nano/inference_vllm_pipeline.py` — `SamplingParams(..., repetition_penalty=1.3)`.
- `funasr/models/fun_asr_nano/inference_vllm_streaming.py` — `SamplingParams(..., repetition_penalty=1.3)`.

### Fix

Centralize the rule in a small, dependency-free helper, `fun_asr_nano/vllm_utils.resolve_repetition_penalty()`, which forces the neutral value (`1.0`) whenever the request runs in prompt-embeds mode and warns **once** so the behavior is discoverable. Every Fun-ASR-Nano `SamplingParams` construction now routes through it, so the failure mode is impossible regardless of the call site. The pipeline and streaming paths additionally now honor a caller-supplied `repetition_penalty` (sanitized) instead of silently ignoring it.

This keeps all serving paths consistent with the already-safe batch `generate()` entry.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [x] Runtime or deployment

## Validation

- [x] `python -m compileall funasr tests` (the new/edited files compile; the only error is the pre-existing Triton-DSL file `funasr/models/sense_voice/whisper_lib/triton_ops.py`, unrelated to this change)
- [x] `python -m unittest tests.test_fun_asr_nano_repetition_penalty` — new tests cover: neutral pass-through, `None` handling, clamping `1.3 → 1.0` in prompt-embeds mode, pass-through for regular token prompts, and warn-once behavior. The helper is dependency-free, so the tests run **without a GPU or vLLM**.

### Failure mode / reproduction

Before this change, starting the server and transcribing audio that reaches the vLLM engine:

```bash
python serve_vllm.py --model FunAudioLLM/Fun-ASR-Nano-2512 --gpu-memory-utilization 0.5 --port 8899
```

aborts the engine with the CUDA scatter assertion above (most visible on longer audio, which produces more decode steps). After this change the penalty is neutralized for prompt-embeds requests and inference completes.

## User impact

Anyone deploying Fun-ASR-Nano via the OpenAI-compatible/REST server, the pipeline path, or the streaming path — including agent integrations that call the transcription endpoint — no longer hits the CUDA crash. The supported batch path is unaffected (it already used `1.0`).

## Notes for reviewers

- No behavior change for callers that already used `repetition_penalty=1.0` (the default).
- The only callers affected are those that previously **crashed**, so there is no working path that regresses.
- The helper is intentionally standard-library only so it imports and tests cleanly without a CUDA device.
- If you prefer the penalty to remain configurable for a future non-embeds decode path, `resolve_repetition_penalty(value, prompt_embeds=False)` passes the value through unchanged.

### DIFF
--- a/funasr/bin/_server_app.py
+++ b/funasr/bin/_server_app.py
@@ -132,8 +132,10 @@ def create_app(device: str = "cuda", preload_model: str = "auto") -> FastAPI:
         if not seg_audios:
             return {"text": "", "segments": [], "duration": len(audio_data)/sr}
 
-        # vLLM generate with repetition_penalty
-        gen_kwargs = {"max_new_tokens": 500, "repetition_penalty": 1.3}
+        # repetition_penalty is left at the neutral 1.0: the Fun-ASR-Nano vLLM
+        # engine runs in prompt-embeds mode, where any other value crashes the
+        # CUDA kernel (see issue #2948 and fun_asr_nano.vllm_utils).
+        gen_kwargs = {"max_new_tokens": 500, "repetition_penalty": 1.0}
         if language:
             gen_kwargs["language"] = language
         if hotwords:
--- a/funasr/models/fun_asr_nano/inference_vllm.py
+++ b/funasr/models/fun_asr_nano/inference_vllm.py
@@ -527,6 +527,8 @@ class FunASRNanoVLLM:
         except ImportError:
             from vllm.inputs.data import EmbedsPrompt
 
+        from funasr.models.fun_asr_nano.vllm_utils import resolve_repetition_penalty
+
         if isinstance(inputs, (str, np.ndarray, torch.Tensor)):
             inputs = [inputs]
 
@@ -535,7 +537,8 @@ class FunASRNanoVLLM:
             temperature=temperature,
             top_p=top_p,
             top_k=top_k if top_k > 0 else -1,
-            repetition_penalty=repetition_penalty,
+            # Prompt-embeds mode has no token IDs to penalize; see #2948.
+            repetition_penalty=resolve_repetition_penalty(repetition_penalty),
             skip_special_tokens=True,
         )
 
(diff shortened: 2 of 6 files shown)


LauraGPT (approved):
Thanks @SuperMarioYL — great fix, and the root-cause analysis is spot on. 🙌

In prompt-embeds mode the request carries no prompt token IDs, so vLLM applying `repetition_penalty` via a scatter over the (empty) prompt token-id tensor is exactly what triggers the CUDA index-out-of-bounds abort. Forcing the penalty back to the neutral `1.0` in that mode — while passing the requested value through for regular token-prompt decoding — is the right call. I especially appreciate that you:

- factored it into a dependency-free `resolve_repetition_penalty()` helper,
- applied it consistently across the server / inference / pipeline / streaming paths,
- and added a CUDA-free unit test.

I checked out the branch and ran `tests/test_fun_asr_nano_repetition_penalty.py` locally — **all 6 tests pass**, and the helper behaves as documented (embeds: `1.3 → 1.0` with a single warning; token mode: `1.3` passthrough; `None → 1.0`). This also resolves #2948.

Merging — thanks for the contribution! 🎉

PR #2988, opened 2026-06-18
Add llama.cpp / GGUF runtime (Fun-ASR-Nano, SenseVoice, Paraformer)

LauraGPT:
## FunASR on llama.cpp / GGUF

Adds a `runtime/llama.cpp/` directory that runs FunASR models entirely on the
llama.cpp / ggml stack — CPU / edge, single binary, no Python at runtime, quantized
GGUF weights. Like whisper.cpp for Whisper; complements the existing PyTorch / ONNX /
vLLM paths for the no-GPU / on-device / embedded case.

### Models (all validated against the PyTorch reference)
| model | architecture | result |
|---|---|---|
| **Fun-ASR-Nano** | SAN-M encoder + adaptor + Qwen3-0.6B LLM (audio embeds injected via `llama_decode` embedding input, llava/mtmd style) | encoder cosine 1.0; end-to-end CER within 0.02% of PyTorch under identical conditions |
| **SenseVoiceSmall** | SAN-M encoder + CTC | CTC token ids identical to PyTorch (108/108); text matches AutoModel exactly |
| **Paraformer** | SAN-M encoder + CIF predictor + SAN-M decoder (non-autoregressive) | decoded text identical to AutoModel; CIF token count exact (105/105) |

All three share a common ggml SAN-M encoder / FSMN / cross-attention and a
kaldi-compatible fbank front end (80-mel, LFR 7/6). Each ships a GGUF export script,
a SentencePiece/tokens detok helper where needed, and a README with the architecture
diagram, build/convert/run quickstart, validation numbers and gotchas.

### Notes
- **Adds a new `runtime/llama.cpp/` directory only — no existing code is modified.**
- WAV input currently assumes 16 kHz mono PCM16; Fun-ASR-Nano supports `--chunk` for
  long audio (a proper VAD front end is on the roadmap).
- The llama.cpp example sources are drop-in (`examples/<name>/`); build instructions
  are in each README.

Please review; not for merge yet.

### DIFF
--- a/runtime/llama.cpp/DESIGN.md
+++ b/runtime/llama.cpp/DESIGN.md
@@ -1,0 +1,337 @@
+# FunASR on llama.cpp / GGUF — Design Document
+
+This document describes the design of the `runtime/llama.cpp` directory: a C++ /
+ggml runtime that runs FunASR models (Fun-ASR-Nano, SenseVoiceSmall, Paraformer)
+without PyTorch, on CPU and edge devices, with quantized GGUF weights. It is the
+counterpart of [whisper.cpp](https://github.com/ggml-org/whisper.cpp) for FunASR.
+
+It is written to be read without the source: it explains *why* the runtime exists,
+*how* each model maps onto ggml, the GGUF weight format, the numerical-fidelity and
+validation methodology, the non-obvious gotchas discovered during the port, and the
+roadmap.
+
+> This is the **shared design document** for the FunASR-on-llama.cpp effort and is
+> kept identical across the FunASR family repos (modelscope/FunASR, Fun-ASR,
+> SenseVoice). The three models share one ggml SAN-M encoder / FSMN / fbank
+> foundation, so the design is documented once here in full; a single-model repo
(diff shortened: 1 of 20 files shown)


LauraGPT:
Thanks for the review. All findings addressed in the latest commit:

- **WAV reader (critical, ×3)** — now validates channel count and bit depth; non-16-bit / zero-channel WAVs are rejected with a clear error instead of dividing by zero. Verified: a malformed 8-bit / `ch=0` file now exits cleanly with `only 16-bit PCM WAV ...` instead of crashing.
- **chunk size (high)** — `chunk_n` is clamped to `>= 1`, so a tiny `--chunk` can no longer cause an infinite loop.
- **fopen output (high)** — output `fopen` is null-checked in funasr-encoder.
- **resource/leak (medium, ×7)** — `llama_context` null-check; `fclose` on all `fread` error paths; `ggml_free`/`gallocr_free`/`backend_free` of contexts before exit.

All three models re-built and re-tested after the changes (Fun-ASR-Nano, SenseVoice CTC ids, Paraformer text) — outputs unchanged / still match the PyTorch reference.

PR #2992, opened 2026-06-19
docs: CPU benchmark vs whisper.cpp (Chinese ASR)

LauraGPT:
## CPU benchmark: FunASR vs whisper.cpp (Chinese)

Adds `runtime/llama.cpp/BENCHMARKS.md` — a head-to-head of the FunASR llama.cpp
runtime against whisper.cpp on the same machine (CPU, 8 threads) and the same
Mandarin audio.

**Headline (60-clip subset, CER lower = better, speed = ×real-time):**

| system | CER | speed |
|---|---|---|
| Fun-ASR-Nano | **5.71%** | LLM decode |
| SenseVoiceSmall | **7.20%** | **19.8×** |
| Paraformer | **9.18%** | **20.9×** |
| whisper.cpp base | 22.96% | 12.7× |
| whisper.cpp small | 17.46% | 4.8× |
| whisper.cpp large-v3-turbo | 16.33% | 3.0× |

On Chinese, the FunASR family is **2.3–2.9× more accurate** and (for the
encoder-only models) **4–6.6× faster** than whisper.cpp at comparable or smaller
size. The doc includes full methodology, a qualitative example, the architectural
reason (non-autoregressive vs autoregressive), and fair-use caveats (this is a
Chinese benchmark; Whisper is a general multilingual model).

Docs-only; adds one file under `runtime/llama.cpp/`.

### DIFF
--- a/runtime/llama.cpp/BENCHMARKS.md
+++ b/runtime/llama.cpp/BENCHMARKS.md
@@ -1,0 +1,100 @@
+# FunASR (llama.cpp / GGUF) vs whisper.cpp — CPU benchmark
+
+How does the FunASR llama.cpp runtime compare with [whisper.cpp](https://github.com/ggml-org/whisper.cpp),
+the de-facto on-device ASR runtime, on **Chinese** speech? This page reports a
+head-to-head on identical hardware and audio.
+
+**TL;DR — for Chinese ASR on CPU, FunASR is more accurate, faster, and comparable
+or smaller in size than whisper.cpp at every model tier.**
+
+## Results
+
+Same machine, **CPU only, 8 threads**, greedy decoding. Accuracy = character error
+rate (CER, lower is better); speed = real-time factor `compute_time / audio_duration`
+(higher `×` is faster, model-load time excluded); whisper forced to Chinese (`-l zh`).
+
+| system | model size (f16 / f32) | **CER** ↓ | **speed** ↑ |
+|---|---|---|---|
+| **FunASR Fun-ASR-Nano** (encoder+adaptor+Qwen3-0.6B) | 0.9 + 0.8 GB | **5.71 %** | LLM decode¹ |
+| **FunASR SenseVoiceSmall** | **449 MB** / 893 MB | **7.20 %** | **19.8× real-time** |
+| **FunASR Paraformer** | **401 MB** / 824 MB | **9.18 %** | **20.9× real-time** |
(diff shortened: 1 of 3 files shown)


LauraGPT:
**Maintainer review (repo admin / benchmark owner).** Great direction — a CPU benchmark vs whisper.cpp on Chinese is exactly the differentiator we should highlight, and "FunASR ≫ whisper on Mandarin CPU" is a real, defensible message. A few things to address before merge:

### 1. CER consistency (blocking)
The table reports Fun-ASR-Nano **5.71%**, SenseVoice **7.20%**, Paraformer **9.18%**. But our **canonical** numbers (184-file benchmark, **micro-average CER + `normalize_zh`**, published on funasr.com / the Fun-ASR README and used everywhere) are **Fun-ASR-Nano 8.06%, SenseVoice 7.81%, Paraformer 10.18%**. Every number here is lower — and Fun-ASR-Nano 5.71% is even below the llama.cpp port's own recorded micro-CER (~9.51% chunked / ~8.2% with VAD).

This looks like a **different, easier test set** (the doc says "FunASR clips fit one segment" → short single-segment clips, no VAD — unlike the canonical 184-file VAD-segmented set). That's fine for a *relative* comparison, but publishing absolute numbers that **contradict our own published CER** invites "your own docs disagree (5.71 vs 8.06)". Please either:
- **(preferred)** re-run on the **canonical 184-file set with micro-CER + `normalize_zh`**, and run whisper.cpp on the *same* set — so the FunASR numbers match what we publish elsewhere and it's apples-to-apples; **or**
- at minimum, **explicitly disclose the test set** (how many clips, which dataset, that they're short single-segment clips differing from the VAD-segmented canonical numbers) and add a one-line note pointing to the canonical numbers, so there's no apparent contradiction.

### 2. Include the reproduction scripts
The doc says `agg_bench.py` + per-system runners "accompany this benchmark" but they aren't in the PR (also flagged by gemini). Please add them.

### 3. Size-column header (gemini)
Header says `f16 / f32` but Fun-ASR-Nano's `0.9 + 0.8 GB` mixes an f32 encoder + f16 LLM — clarify.

The whisper.cpp tiers (22.96 / 17.46 / 16.33) and the relative ranking are great. Once the FunASR absolute CER is consistent with our published口径 (or the test set is clearly disclosed) + scripts included, I'll merge — this will be strong promotion material. Same applies to the mirror PRs Fun-ASR#117 / SenseVoice#294.

PR #2993, opened 2026-06-19
fix(glm_asr): warn when vLLM dtype=fp16 (degraded output)

SuperMarioYL:
### What & why

The GLM-ASR vLLM engine (`GLMASRVLLMEngine`) accepts `dtype="fp16"`, but — exactly like Fun-ASR-Nano — fp16 can produce **degraded or garbage transcription** due to numerical overflow in the audio embedding path.

Fun-ASR-Nano already guards against this: `inference_vllm.py` logs a warning when `dtype` resolves to `float16`, pointing users at `bf16`/`fp32` (added in #2980). The newer GLM-ASR engine has the **same** audio-tower-in-torch + LM-in-vLLM architecture and the same `dtype_map` default, but never got that warning — so a user who passes `fp16` gets silently poor output with no hint about what went wrong.

This PR closes that gap.

### Change

- Add `funasr/models/glm_asr/vllm_utils.py` with `warn_if_degraded_dtype()` — a one-time warning when `fp16` is requested. The requested value is **still honoured** (some GPUs only support fp16); the user is just told once why output may be poor.
- Wire it into `GLMASRVLLMEngine.__init__`. Behaviour for every dtype is otherwise unchanged — `self.torch_dtype = dtype_map.get(warn_if_degraded_dtype(dtype), torch.bfloat16)` keeps the existing default.

Fun-ASR-Nano emits this warning inline; I put it in a tiny helper so it can be unit-tested without a GPU/vLLM, matching the dependency-free `fun_asr_nano/vllm_utils.py` pattern already used for the prompt-embeds repetition-penalty guard (#2974). Happy to inline it instead if you'd prefer it to match the sibling line-for-line.

### Tests

`tests/test_glm_asr_vllm_dtype.py` — covers fp16 warns (once), safe dtypes don't warn, and the value is returned unchanged. The helper is stdlib-only; tests run without a GPU or vLLM (Python 3.8+ compatible, no `assertNoLogs`).

```
$ python -m pytest tests/test_glm_asr_vllm_dtype.py -v
4 passed
```

### DIFF
--- a/funasr/models/glm_asr/inference_vllm.py
+++ b/funasr/models/glm_asr/inference_vllm.py
@@ -101,8 +101,10 @@ class GLMASRVLLMEngine:
         from vllm import LLM
         from transformers import AutoProcessor, AutoConfig, AutoModel as HFAutoModel
 
+        from funasr.models.glm_asr.vllm_utils import warn_if_degraded_dtype
+
         self.device = device
-        self.torch_dtype = dtype_map.get(dtype, torch.bfloat16)
+        self.torch_dtype = dtype_map.get(warn_if_degraded_dtype(dtype), torch.bfloat16)
         self.model_dir = model_dir
 
         logger.info(f"Loading GLM-ASR audio components from {model_dir}")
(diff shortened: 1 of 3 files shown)


LauraGPT:
Thanks @SuperMarioYL — clean fix with a dependency-free unit test, and the fp16-degradation guard mirrors the known Fun-ASR-Nano behavior. Merging. 🙏

PR #3014, opened 2026-06-22
fix(glm_asr): dedup result keys for duplicate audio basenames

SuperMarioYL:
### Problem

`GLMASRVLLMEngine.generate()` derives each result key from the audio file basename:

```python
key = os.path.splitext(os.path.basename(inputs[i]))[0] if isinstance(inputs[i], str) else f"sample_{i}"
```

So two inputs that live in different directories but share a basename — a very common batch-transcription layout, e.g. `spk1/segment.wav` and `spk2/segment.wav`, or `meeting/01/audio.wav` and `meeting/02/audio.wav` — both map to the same key (`"segment"` / `"audio"`). As soon as the returned list is folded into the canonical `{r["key"]: r["text"]}` mapping, every collision silently overwrites the previous transcript: the caller gets **fewer results than inputs, with no error and no warning**.

### Fix

Add a small module-level `_dedup_keys` helper that appends a deterministic `_N` suffix to later collisions while preserving order and first-occurrence names (`seg`, `seg_1`, `seg_2`, …), and wire it into `generate()`. The suffix search skips keys that are already taken, so a result can never collide even when an input is literally named `seg_1.wav` alongside two `seg.wav`. A one-time `logger.warning` tells the user when basenames clashed so the behaviour is discoverable.

No behaviour change for the common case (all-distinct basenames return exactly the keys they did before).

### Tests

`tests/test_glm_asr_vllm_dedup_keys.py` covers the pure helper (order preservation, deterministic suffixes, the `seg`/`seg_1`/`seg` re-collision edge case, empty input, no input mutation) and the end-to-end `generate()` path via a `sys.modules` vLLM stub — no GPU, vLLM, or model download needed.

```
$ python -m pytest tests/test_glm_asr_vllm_dedup_keys.py -q
9 passed
```

The helper is kept local to `glm_asr/inference_vllm.py` (matching the existing inline helper style) so the change is fully self-contained.

### DIFF
--- a/funasr/models/glm_asr/inference_vllm.py
+++ b/funasr/models/glm_asr/inference_vllm.py
@@ -80,6 +80,63 @@ def prepare_glmasr_vllm_dir(model_dir: str) -> str:
     return output_dir
 
 
+# Warn only once per process so batch loops do not spam the log.
+_warned_dup_keys = False
+
+
+def _dedup_keys(keys):
+    """Make result keys unique while preserving order and first-occurrence names.
+
+    Each result key is derived from the audio file basename
+    (``os.path.splitext(os.path.basename(path))[0]``), so two inputs that live
+    in different directories but share a basename -- e.g. ``spk1/segment.wav``
+    and ``spk2/segment.wav`` -- both map to ``"segment"``. A downstream
+    ``{r["key"]: r["text"]}`` mapping (the canonical FunASR result shape) would
+    then silently drop all but the last colliding entry, returning fewer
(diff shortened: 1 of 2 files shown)


LauraGPT:
Thanks for the careful fix and the GPU-free unit tests 🙏 The silent-drop on colliding basenames in `{r["key"]: r["text"]}` is a real correctness bug, and the deterministic `_N` suffix (plus the guard against suffixes re-colliding) is exactly right. Ran the test suite locally — all 9 cases pass. Merging.

PR #3041, opened 2026-06-29
fix: keep Fun-ASR-Nano vLLM repetition penalty neutral

LauraGPT:
## Summary
- keep Fun-ASR-Nano vLLM repetition penalty neutral by default in server, pipeline, and streaming paths
- allow pipeline/streaming callers to opt into `repetition_penalty` explicitly
- update vLLM guides to remove the outdated `repetition_penalty=1.3` hardcoding note

## Why
Fun-ASR-Nano uses prompt-embeds with vLLM. A non-neutral repetition penalty can destabilize repeated API calls / segmented decoding in some vLLM environments, and users in #3031 are currently debugging repeated output from an API wrapper. The base `AutoModelVLLM.generate()` path already defaults to `repetition_penalty=1.0`; this PR aligns the remaining service and helper paths with that safer default.

## Verification
- `git diff --check`
- `python3 -m py_compile funasr/bin/_server_app.py funasr/models/fun_asr_nano/inference_vllm_pipeline.py funasr/models/fun_asr_nano/inference_vllm_streaming.py`
- AST parse of the three changed Python files
- `rg` scan confirmed no remaining docs/server hardcoding of `repetition_penalty=1.3`

Refs #3031

### DIFF
--- a/docs/vllm_guide.md
+++ b/docs/vllm_guide.md
@@ -307,7 +307,7 @@ for result in engine.streaming_generate("audio.wav", language="中文"):
 | 1.5–3.0 s | Partially correct |
 | > 3.0 s | Accurate output |
 
-> Note: `repetition_penalty=1.3` is hardcoded internally to prevent short-chunk repetition degradation.
+> Note: `repetition_penalty` defaults to `1.0` for Fun-ASR-Nano vLLM paths. Tune it explicitly only after validating that it improves your workload.
 
 ---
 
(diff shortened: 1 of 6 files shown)


LauraGPT:
Closing this as superseded by the current main branch. main already contains the safer implementation via funasr.models.fun_asr_nano.vllm_utils.resolve_repetition_penalty, keeps the server path neutral at repetition_penalty=1.0, and documents that EmbedsPrompt should not use non-neutral repetition penalties. I will point #3031 at the main-branch fix instead of duplicating it here.

PR #3105, opened 2026-07-02
Migrate obsoleted editdistance to rapidfuzz

excosy:
## Summary

Since `editdistance` has been archived and cannot be newly installed on Windows, replace it with `rapidfuzz`.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [x] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [x] Docs or links checked

### DIFF
--- a/funasr/metrics/common.py
+++ b/funasr/metrics/common.py
@@ -9,9 +9,9 @@
 import json
 import logging
 import sys
+from itertools import groupby
 
 from rapidfuzz.distance import Levenshtein
-from itertools import groupby
 import numpy as np
 import six
 
@@ -222,7 +222,7 @@ class ErrorCalculator(object):
             ref_chars = seq_true_text.replace(" ", "")
             char_eds.append(Levenshtein.distance(hyp_chars, ref_chars))
             char_ref_lens.append(len(ref_chars))
-        return float(sum(char_eds)) / sum(char_ref_lens)
+        return float(sum(char_eds)) / sum(char_ref_lens) if char_eds else None
 
     def calculate_wer(self, seqs_hat, seqs_true):
         """Calculate sentence-level WER score.
@@ -240,4 +240,4 @@ class ErrorCalculator(object):
             ref_words = seq_true_text.split()
             word_eds.append(Levenshtein.distance(hyp_words, ref_words))
             word_ref_lens.append(len(ref_words))
-        return float(sum(word_eds)) / sum(word_ref_lens)
+        return float(sum(word_eds)) / sum(word_ref_lens) if word_eds else None
--- a/funasr/metrics/common.py
+++ b/funasr/metrics/common.py
@@ -10,6 +10,7 @@ import json
 import logging
 import sys
 
+from rapidfuzz.distance import Levenshtein
 from itertools import groupby
 import numpy as np
 import six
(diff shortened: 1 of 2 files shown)


LauraGPT:
Thanks for migrating away from archived `editdistance`; the direction makes sense, especially for Windows installs.

I validated current head `71af8eba` locally and found one remaining issue from the high-priority review feedback: `calculate_cer()` / `calculate_wer()` still divide by zero when there is at least one hypothesis but every reference is empty or only spaces. The current `if char_eds else None` guard only handles empty input lists, not zero total reference length.

I prepared a minimal TDD patch on `LauraGPT/FunASR:codex/fix-pr-3105-empty-ref-metrics`:

- adds regression coverage for empty-reference CER/WER
- keeps normal Levenshtein semantics for non-empty references
- returns `None` when total reference length is zero, matching `calculate_cer_ctc()`'s no-valid-reference behavior

Validation on ind-gpu8:

- `python -m pytest tests/test_metrics_common_rapidfuzz.py -q` -> 2 passed
- `python -m py_compile funasr/metrics/common.py tests/test_metrics_common_rapidfuzz.py setup.py` -> OK
- `git diff --check` -> OK
- temporary real `rapidfuzz>=3.0.0` API probe confirms string and token-list distance semantics

Compare: https://github.com/modelscope/FunASR/compare/main...LauraGPT:FunASR:codex/fix-pr-3105-empty-ref-metrics?expand=1

LauraGPT (changes requested):
Thanks for taking on the `editdistance` -> `rapidfuzz` migration. The dependency direction is good, but current head `71af8eba` still has a correctness blocker before merge.

`calculate_cer()` and `calculate_wer()` still divide by zero when the input list is non-empty but every reference is empty or whitespace-only. The current guard only checks whether there are edit-distance entries, not whether total reference length is zero.

I reproduced it with a focused regression test and prepared a small fix branch:

- branch: `LauraGPT/FunASR:codex/fix-pr-3105-empty-ref-metrics`
- compare: https://github.com/modelscope/FunASR/compare/main...LauraGPT:FunASR:codex/fix-pr-3105-empty-ref-metrics?expand=1

Validation on ind-gpu8:

- red before fix: `python -m pytest tests/test_metrics_common_rapidfuzz.py -q` -> `ZeroDivisionError`
- green after fix: `python -m pytest tests/test_metrics_common_rapidfuzz.py -q` -> 2 passed
- `python -m py_compile funasr/metrics/common.py tests/test_metrics_common_rapidfuzz.py setup.py` -> OK
- `git diff --check` -> OK
- temporary real `rapidfuzz>=3.0.0` API probe confirmed string and token-list distance semantics

Please adopt the zero-reference guard before this PR is merged. Returning `None` when total reference length is zero matches the existing CTC metric behavior.

LauraGPT:
Thanks again for starting the `editdistance` -> `rapidfuzz` migration.

I merged the maintainer-owned replacement in #3114 so we could include the empty-reference CER/WER guard before changing the dependency on `main`.

Your PR was the basis for the migration; the replacement keeps the same dependency direction and adds the missing zero-reference behavior:

- `calculate_cer(["abc"], ["   "])` returns `None`
- `calculate_wer(["hello world"], ["   "])` returns `None`

I am closing this PR to avoid having two open PRs for the same migration.

PR #3220, opened 2026-07-14
Fix MCP model and language configuration

LauraGPT:
## Summary

- honor FUNASR_MODEL and forward the MCP language hint to AutoModel.generate
- validate audio paths and supported language values, expand home-directory paths, and return MCP tool errors instead of terminating the stdio server
- align the tool contract and README with default SenseVoiceSmall capabilities and report the installed FunASR version dynamically

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Example or demo
- [x] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [x] python -m compileall funasr examples tests
- [x] Docs and capability claims checked against the implementation and a real model result
- [x] Runtime command tested

Additional checks:
- 13 focused pytest checks passed
- 9 stdlib MCP tests passed on Python 3.8.20
- Python 3.7 syntax parsing and Ruff target-version py37 passed
- official MCP SDK 1.27.1 initialize, tools/list, and error-path tools/call passed
- real FunASR 1.3.14 CPU inference passed on a 45.1-second audio file using an explicit zh language hint and FUNASR_MODEL local path

## User impact

Agent builders can now select the documented model, provide an effective language hint, use paths beginning with ~, and recover from inference errors without restarting the MCP server. Directory metadata no longer advertises unsupported default capabilities.

## Notes for reviewers

The default remains iic/SenseVoiceSmall with fsmn-vad. Speaker fields and segment timestamps are still preserved when a configured model returns sentence_info; they are simply no longer advertised as default SenseVoiceSmall capabilities.

Docker is unavailable in the ind-gpu8 Kubernetes task environment, so the image build was not rerun. The same Python entrypoint passed stdlib, official SDK, and real-model tests.

### DIFF
--- a/examples/mcp_server/README.md
+++ b/examples/mcp_server/README.md
@@ -1,6 +1,6 @@
 # FunASR MCP Server
 
-[Model Context Protocol](https://modelcontextprotocol.io/) server that gives AI assistants the ability to transcribe audio.
+[Model Context Protocol](https://modelcontextprotocol.io/) server that gives AI assistants local audio transcription with SenseVoiceSmall by default.
 
 ## Setup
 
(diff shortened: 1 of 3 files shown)


LauraGPT (commented):
Maintainer validation at head 3a9813535 after syncing current main:
- 13 focused MCP tests passed
- 9 stdlib tests passed on Python 3.8.20; Python 3.7 syntax gate passed
- Ruff py37, full compileall, and diff checks passed
- official MCP SDK 1.27.1 initialize and tools/list passed
- real FunASR 1.3.14 CPU inference passed with an explicit zh hint and FUNASR_MODEL local path

Gemini reported no review findings on the implementation commit; the only later change is the clean merge of reviewed PR #3219 from main.

PR #3240, opened 2026-07-16
fix: disable duplicate dynamic silence logic in DynamicStreamingVAD

LiangRx0501:
## Summary

Disable the model-level dynamic silence logic when `DynamicStreamingVAD` manages the threshold through `_apply_dynamic_threshold()`.

This prevents `model.py` from overwriting the caller-configured VAD threshold on every streaming chunk.

### DIFF
--- a/funasr/models/fsmn_vad_streaming/dynamic_vad.py
+++ b/funasr/models/fsmn_vad_streaming/dynamic_vad.py
@@ -134,7 +134,7 @@ class DynamicStreamingVAD:
         res = self.model.generate(
             input=[audio_chunk], cache=self.cache,
             is_final=is_final, chunk_size=self.chunk_size_ms,
-            silence_schedule=self.silence_schedule,
+            silence_schedule=self.silence_schedule, #fix here
         )
 
         signals = res[0].get("value", [])
--- a/funasr/models/fsmn_vad_streaming/dynamic_vad.py
+++ b/funasr/models/fsmn_vad_streaming/dynamic_vad.py
@@ -134,6 +134,7 @@ class DynamicStreamingVAD:
         res = self.model.generate(
             input=[audio_chunk], cache=self.cache,
             is_final=is_final, chunk_size=self.chunk_size_ms,
+            silence_schedule=self.silence_schedule,
         )
 
         signals = res[0].get("value", [])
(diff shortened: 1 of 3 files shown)


LauraGPT (changes requested):
Thanks for fixing the duplicate threshold controller. I confirmed on the real CPU FSMN-VAD that `dynamic_silence=False` prevents the model-level schedule from overwriting `DynamicStreamingVAD` after the cache exists: a custom 3000 ms schedule now remains at a 2850 ms counter threshold instead of being reset to 1850 ms.

There is still a first-call correctness gap that needs to be fixed before merge. `_apply_dynamic_threshold()` runs before `self.model.generate()`, so the first call has no `cache["stats"]` to update. With this patch disabling the model schedule, `generate()` initializes the first call with the model's default 800 ms end-silence setting, regardless of the wrapper's configured schedule. Because `feed()` explicitly accepts an arbitrary-length chunk and the model internally slices it, this makes segmentation depend on how the caller chunks identical audio.

Real-model reproduction at `ad3f3e402`, using the repository's `runtime/funasr_api/asr_example.wav` plus two seconds of silence and `silence_schedule=[(inf, 10000)]`:

- one `feed(full_audio)`: returns `[[610, 5440]]`; counter threshold is `650` (800 ms total);
- `feed(first_60_ms)` followed by `feed(rest)`: returns `[]`; counter threshold is `9850` (10000 ms total).

Please initialize the first model call with the wrapper's desired end-silence setting (for example through the existing `max_end_silence_time` initialization path, or an equivalent cache initialization) while keeping the model-level dynamic schedule disabled.

Please also add focused regression coverage for both parts of the contract:

1. the wrapper's custom silence schedule and `speech_noise_thres` are not overwritten by the model-level defaults;
2. identical audio produces the same result when supplied as one initial chunk or split after the first streaming chunk.

The current focused suite (`tests/test_fsmn_vad_dynamic_silence.py` plus `tests/test_realtime_ws_service.py`) passes 15 tests, but none exercises the `DynamicStreamingVAD`/model interaction changed by this PR.

LauraGPT (approved):
Verified the first-call fix on head `92b749cdf`.

- The initial FSMN-VAD cache now receives the wrapper-configured end-silence and `speech_noise_thres` values before processing audio.
- The new wrapper/cache regression failed on the previous head (`650 != 9850`, and one-chunk `[[0, 1000]]` versus split `[]`) and passes after the fix.
- The focused suite passes **17/17** tests; compile, targeted Ruff, and diff checks pass.
- A real CPU `fsmn-vad` run with the repository audio plus two seconds of silence produces identical one-chunk and 60 ms-split results (`[]`), with both caches at threshold `9850` and noise threshold `0.73`.

The requested threshold-preservation and chunking-invariance contract is covered.

PR #3299, opened 2026-07-20
feat(server): add --model-path and --hub options to funasr-server

liudonghua123:
- Add --model-path argument for local model or model ID override
- Add --hub argument to specify model hub (ms for ModelScope, hf for HuggingFace)
- Support custom model loading via AutoModel with specified path and hub
- Update API endpoints to handle 'custom' model type

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [x] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [x] Docs or links checked

## User impact

Nothing.


LauraGPT (changes requested):
Thanks for the contribution. I think this needs one more pass before merge: the new custom model path does not appear to load correctly when users run the documented command with only `--model-path`.

In `create_app(..., preload_model="auto", model_path=...)`, the preload branch calls `_load_fallback(preload_model if preload_model != "auto" else "custom")`, but `custom` is not in `FALLBACK_CONFIGS`. That means `_load_fallback("custom")` will not have a config to copy. Later, a request with `model=custom` calls `_process_fallback(preload_model, ...)`, which is still `auto` for the documented command, so it also does not map cleanly to a loaded model.

Could you add a small test or manual validation log for the two intended paths?

1. `funasr-server --model-path /path/to/model --hub ms`
2. `funasr-server --model-path FunAudioLLM/Fun-ASR-Nano-2512 --hub hf --model fun-asr-nano`

A simpler implementation may be to choose the serving id explicitly when `model_path` is set, store the custom model under that id, and make `/v1/models` plus `model=custom` resolve to the same loaded entry.

LauraGPT (approved):
Maintainer follow-up pushed in `1dbf9bfc` and synced with current `main` in `05bc48a9`.

This addresses the blocking server-loader review items:
- default `fun-asr-nano` now keeps using the Hugging Face hub unless a custom `--model-path` is provided;
- AutoModel fallback for default `fun-asr-nano` uses the same hub rule;
- custom model paths can load through an empty fallback config and are cached under the `custom` key.

Validation on `ind-gpu8`:
- `python3 -m py_compile funasr/bin/_server_app.py funasr/bin/server.py tests/test_server_app_openai_segments.py`
- `python3 -m pytest -q tests/test_cli.py tests/test_server_app_openai_segments.py -q` -> 8 passed
- `git diff --check`

PR #3300, opened 2026-07-20
Update funasr-wss-client-2pass.cpp

dio-del:
## Summary

Fixed the issue where the client lost the last segment of audio in online mode and ensured that the server no longer returns offline results in online mode. In online mode, you can see that in the figure, 2pass-online misses the character “型”, and the result can only be obtained using 2pass-offline. Also, offline results should not be returned in online mode.

## Change Type

- [√ ] Bug fix
- [ ] Documentation update
- [ ] Example or demo
- [ ] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Verification

before
before
after
fixed

- [ ] Documentation or links checked
- [√ ] Runtime/deployment commands tested

## User Impact

New users and researchers

## Notes for Reviewers

None

### DIFF
--- a/runtime/websocket/bin/funasr-wss-client-2pass.cpp
+++ b/runtime/websocket/bin/funasr-wss-client-2pass.cpp
@@ -4,655 +4,649 @@
  */
 /* 2022-2023 by zhaomingwork */
 
-// client for websocket, support multiple threads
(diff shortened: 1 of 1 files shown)


dio-del:
fix wrong file

PR #3301, opened 2026-07-20
fix: client lost the last segment of audio in online mode

dio-del:
## Summary

Fixed the issue where the client lost the last segment of audio in online mode and ensured that the server no longer returns offline results in online mode. In online mode, you can see that in the figure, 2pass-online misses the character “型”, and the result can only be obtained using 2pass-offline. Also, offline results should not be returned in online mode.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [ ] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

before
before
after
after

- [ ] Docs or links checked

## User impact

New users and production deployers

## Notes for reviewers

None

### DIFF
--- a/runtime/websocket/bin/websocket-server-2pass.cpp
+++ b/runtime/websocket/bin/websocket-server-2pass.cpp
@@ -65,7 +65,7 @@ context_ptr WebSocketServer::on_tls_init(tls_mode mode,
   return ctx;
 }
 
-nlohmann::json handle_result(FUNASR_RESULT result, websocketpp::connection_hdl& hdl, std::map<websocketpp::connection_hdl, std::shared_ptr<FUNASR_MESSAGE>,std::owner_less<websocketpp::connection_hdl>>& data_map) {
+nlohmann::json handle_result(FUNASR_RESULT result, websocketpp::connection_hdl& hdl, std::map<websocketpp::connection_hdl, std::shared_ptr<FUNASR_MESSAGE>,std::owner_less<websocketpp::connection_hdl>>& data_map, std::string modetype) {
   std::shared_ptr<FUNASR_MESSAGE> data_msg = nullptr;
   auto it = data_map.find(hdl);
   if (it != data_map.end()) {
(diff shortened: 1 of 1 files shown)


LauraGPT (changes requested):
Thanks for the fix and screenshots. I agree with the direction of not returning 2pass-offline text while the client requested online mode, but I think this needs one more small adjustment before merge.\n\nAfter this change, when `modetype == "online"`, the code skips the block that currently advances `data_msg->index` and resets `data_msg->is_sentence_started = false`. That means a final online result can leave the connection in a started-sentence state; the next utterance on the same connection may not emit `slice_type = 0` and the sentence index may not advance.\n\nCould you update the final-result path for online mode so it still closes the sentence state without sending the offline transcript? One possible approach is to pass `is_final` into `handle_result` and, for `modetype == "online" && is_final`, set the final metadata/reset the state separately from the offline-text override.\n\nI also tried to compile `funasr-wss-server-2pass` locally. CMake configuration succeeded, but this environment is missing the generated ONNXRuntime header `onnxruntime_run_options_config_keys.h`, so the full runtime build could not complete here. Please include your C++ build command/output with the update if possible.

LauraGPT (approved):
Fixed the requested final-state handling in `0e8e7376`.

What changed:
- `handle_result()` now receives `is_final` and `const std::string& modetype`.
- `modetype == "online"` still suppresses the `2pass-offline` transcript, but on the final chunk it now advances `data_msg->index` and resets `data_msg->is_sentence_started = false` so the next utterance can start with `slice_type = 0`.
- Added null checks before constructing `std::string` from both `FunASRGetResult()` and `FunASRGetTpassResult()`.

Validation run on the PR branch:

```bash
git diff --check
c++ -std=c++14 -fsyntax-only \
  -Iruntime/websocket/bin \
  -Iruntime/websocket/third_party/websocket \
  -Iruntime/websocket/third_party/asio/asio/include \
  -Iruntime/websocket/third_party/json/include \
  -Iruntime/onnxruntime/include \
  -Iruntime/onnxruntime/src \
  -Iruntime/onnxruntime/third_party/yaml-cpp/include \
  -Iruntime/onnxruntime/third_party/kaldi-native-fbank \
  -Iruntime/onnxruntime/third_party/jieba/include \
  -Iruntime/onnxruntime/third_party/jieba/include/limonp/include \
  -Iruntime/onnxruntime/third_party \
  -Iruntime/onnxruntime/third_party/kaldi \
  -Iruntime/onnxruntime/third_party/openfst/src/include \
  -Iruntime/onnxruntime/third_party/glog/src \
  -Ibuild-websocket-pr3301/glog \
  -Iruntime/onnxruntime/third_party/gflags \
  -Ibuild-websocket-pr3301/gflags/include \
  runtime/websocket/bin/websocket-server-2pass.cpp
```

A full target build was attempted with:

```bash
cmake -S runtime/websocket -B build-websocket-pr3301 \
  -DENABLE_PORTAUDIO=OFF \
  -DONNXRUNTIME_DIR=/cpfs_speech/user/zhifu.gzf/codebase/FunASR-pr3301-fix/runtime/onnxruntime \
  -DFFMPEG_DIR=/usr \
  -DCMAKE_BUILD_TYPE=Release
cmake --build build-websocket-pr3301 --target funasr-wss-server-2pass -j2
```

CMake configuration completed, but the full build still stops before the websocket target at the existing ONNXRuntime header issue:

```text
runtime/onnxruntime/src/precomp.h:31:10: fatal error: onnxruntime_run_options_config_keys.h: No such file or directory
```

PR #3448, opened 2026-07-30
fix(runtime): prevent leaks in two-pass timestamp inference

wangxiuwen:
## What does this PR do?

Fix two ownership leaks in the timestamp pass of `FunTpassInferBuffer`:

- Replace the per-iteration heap allocations for the one-element input pointer and length arrays with stack arrays.
- Release the fetched `AudioFrame` before the empty-result early `continue`.

## Why is this needed?

The input arrays were allocated with `new[]` on every timestamp-pass iteration and were never released. In addition, when `SplitStr` returned no fields, control skipped the existing frame cleanup at the end of the loop.

Both leaks accumulate while processing two-pass streaming audio. The stack arrays have the same lifetime required by the synchronous `Forward` call and require no ownership management.

## Validation

- `git diff --check`
- Verified both `Forward` overloads still receive the same one-element pointer and length arrays.
- Verified every loop exit now releases the fetched frame.

### DIFF
--- a/runtime/onnxruntime/src/funasrruntime.cpp
+++ b/runtime/onnxruntime/src/funasrruntime.cpp
@@ -564,12 +564,8 @@
 			if (wfst_decoder){
 				wfst_decoder->StartUtterance();
 			}
-			float** buff;
-			int* len;
-			buff = new float*[1];
-        	len = new int[1];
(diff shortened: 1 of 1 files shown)


LauraGPT (approved):

PR #3449, opened 2026-07-30
fix(runtime): free custom AVIO buffers on all exits

wangxiuwen:
## What does this PR do?

Free the buffer owned by the custom `AVIOContext` on every exit from the in-memory FFmpeg loader.

The cleanup helper follows FFmpeg's custom-I/O example: close the format input first, free the context's current buffer, and then call `avio_context_free`. It deliberately reads `avio_ctx->buffer` at cleanup time because libavformat may replace the original buffer while probing.

Reference: https://ffmpeg.org/doxygen/trunk/avio_read_callback_8c-example.html

## Why is this needed?

The error paths previously called only `avio_context_free`, which does not release the caller-supplied/current AVIO buffer. Every failed in-memory decode therefore leaked the copied input buffer. The success path also freed the context structure directly instead of using the public FFmpeg cleanup API.

## Validation

- `git diff --check`
- Audited all seven error exits and the success exit in `Audio::FfmpegLoad(const char*, int)`.
- Verified format cleanup precedes custom AVIO cleanup, matching the FFmpeg example.

### DIFF
--- a/runtime/onnxruntime/src/audio.cpp
+++ b/runtime/onnxruntime/src/audio.cpp
@@ -33,6 +33,18 @@ extern "C" {
 using namespace std;
 
 namespace funasr {
+#if !defined(__APPLE__) && defined(ENABLE_FFMPEG)
+namespace {
+void FreeAvioContext(AVIOContext** avio_ctx) {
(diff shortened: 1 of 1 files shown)


LauraGPT (approved):

PR #3450, opened 2026-07-30
fix(runtime): make AudioFrame copies own their sample data

wangxiuwen:
## What does this PR do?

Complete `AudioFrame`'s copy ownership semantics:

- Deep-copy the owned sample buffer in the copy constructor.
- Copy `global_start` and `global_end` metadata.
- Add a deep-copy assignment operator with self-assignment and allocation-failure guards.

## Why is this needed?

`AudioFrame` owns `data` and releases it in its destructor. The existing copy constructor copied only part of the metadata and silently dropped the sample buffer. The implicit copy-assignment operator was more dangerous: it shallow-copied the owned pointer, allowing two frames to free the same allocation.

Implementing both copy operations keeps the existing copyable API while giving each frame independent ownership of its samples.

## Validation

- `git diff --check`
- Verified the copy constructor and assignment operator copy all frame fields.
- Verified self-assignment leaves the frame unchanged.
- Verified assignment allocates the replacement before releasing the destination buffer.
- Verified `AudioFrame` copy operations are not currently used by the in-tree runtime call sites, so this does not change hot-path behavior.

### DIFF
--- a/runtime/onnxruntime/include/audio.h
+++ b/runtime/onnxruntime/include/audio.h
@@ -24,6 +24,7 @@ class AudioFrame {
     AudioFrame();
     AudioFrame(int len);
     AudioFrame(const AudioFrame &other);
+    AudioFrame& operator=(const AudioFrame &other);
     AudioFrame(int start, int end, bool is_final);
 
     ~AudioFrame();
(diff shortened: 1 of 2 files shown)


LauraGPT (approved):

PR #3456, opened 2026-08-02
feat(nano): support LoRA finetuning of the Qwen3 LLM

KamitobiHaru:
## Summary

The FunASR-Nano model config already ships `llm_conf.use_lora` / `llm_conf.lora_conf`
(r, lora_alpha, lora_dropout, target_modules), but **nothing consumed them** — `use_lora` was silently ignored and training fell back to the full-parameter LLM path. This PR wires LoRA up for the Qwen3-0.6B LLM inside `FunASRNano`.

## Changes

- **`funasr/models/fun_asr_nano/model.py`**: when `llm_conf.use_lora` is true, replace the LLM target `nn.Linear` layers (default `q_proj`/`v_proj`) with `lora.Linear` adapters (reusing the existing `funasr/models/lora/` module used by Paraformer). The pretrained base weight is shared and frozen; trainable `lora_A`/`lora_B` are added. Adapter params are created in the base weight's dtype (bf16), so the LoRA path works without relying on
 autocast to reconcile dtypes.
- **`examples/.../fun_asr_nano/lora_finetune.sh`** + **`docs/finetune.md` / `finetune_zh.md`**: LoRA recipe with freeze semantics and deployment (fold) notes.

## Design notes

- Pairs with the existing `lora_only` / `mark_only_lora_as_trainable` flow already in
 `train.py` / `train_ds.py`:
 - `lora_only=true` (default in the example) freezes every non-LoRA parameter → pure-LoRA.
 - `lora_only=false` + unfreeze encoder/adaptor conf keeps them trainable while LoRA-tweaking only the LLM (the recipe used for the medical ASR deliverable).
- Checkpoints save the full state dict (unchanged base + adapter params), so a LoRA run can be resumed / decoded with the same `use_lora=true` config. No new save/load code.
- Inference needs no merge step (forward adds the adapter output); for a standalone deployment checkpoint, fold with `W' = W + (alpha/r) * B @ A` and drop the lora keys.

## Validation

- Real model build: 56 adapters injected (28 Qwen3 layers × q_proj+v_proj), base frozen,
 LoRA trainable.
- Optimizer step updates only the 112 LoRA params (0 non-LoRA change).
- 2-step smoke training via `funasr-train-ds` completes; saved checkpoint's `lora_A`/`lora_B` are bf16 matching the base dtype; merge folds all 56 pairs cleanly.
- `eval()`/`train()` auto merge/unmerge verified (matches the upstream `lora.Linear`
 convention used by Paraformer).

### DIFF
--- a/examples/industrial_data_pretraining/fun_asr_nano/docs/finetune.md
+++ b/examples/industrial_data_pretraining/fun_asr_nano/docs/finetune.md
@@ -93,6 +93,46 @@ bash finetune.sh
 - For training data less than 5000 hours, it is recommended to fine-tune the audio_encoder and audio_adaptor.
 - For training data greater than 10000 hours, it is recommended to perform full-parameter fine-tuning.
 
+## LoRA Finetune
+
+As an alternative to full-parameter / partial fine-tuning, you can LoRA-tune the
+Qwen3-0.6B LLM (attaching adapters to its `q_proj` / `v_proj` linear layers):
+
+```
+bash lora_finetune.sh
+```
+
+What it sets:
+
+- `llm_conf.use_lora=true` — injects `LoRALinear` adapters into the target layers
+  of the LLM (base weights shared and frozen, trainable `lora_A` / `lora_B` added).
+- `lora_only=true` — freezes every non-LoRA parameter (audio encoder, adaptor and
+  CTC decoder), so only the adapters train. Checkpoints still save the full state
+  dict (base weights unchanged + adapter params), so you can resume or decode with
+  the same `use_lora=true` config.
+- `llm_conf.freeze=true` — keeps the LLM base weights frozen (redundant with
+  `lora_only`, but explicit).
+
+LoRA hyper-parameters live under `llm_conf.lora_conf` (defaults ship with the
+model config): `r`, `lora_alpha`, `lora_dropout`, `target_modules`. Override them
+on the command line, e.g.:
+
+```
+++llm_conf.lora_conf.r=32
+```
+
+To keep the audio encoder / adaptor **trainable** while LoRA-tweaking only the LLM
+(a good middle ground for a few hundred hours of domain data), set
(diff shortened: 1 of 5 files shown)


LauraGPT (changes requested):
The implementation passed my focused local contract check on commit `11089066ba6a401fc52ecc2f647e27558d65d25b`: target-only replacement, unchanged initial forward output, frozen base/trainable adapter parameters, train/eval transitions, and a full state-dict round trip all behaved correctly. The existing Nano regression set also passes (`27 passed`).

However, none of the 225 added lines includes repository test coverage for the new LoRA path. Please add a small offline unit test (a tiny fake LLM is sufficient; no model download is needed) that covers:

1. only configured target modules are replaced;
2. the initial output matches the original linear layers;
3. base weights stay frozen while `lora_A`/`lora_B` remain trainable with the documented configuration;
4. a non-zero adapter survives a state-dict save/load round trip and train/eval transitions;
5. the no-matching-target configuration has an explicit tested outcome.

Once equivalent coverage is committed, this should be quick to re-review.

LauraGPT (approved):
Re-reviewed exact head cd1ab320. The requested committed regression coverage is now present and meaningfully red on the pre-feature base (7/7 fail there).\n\nVerification on this head:\n- LoRA injection test file: 7/7 passed; Nano-focused suite passed.\n- Real cached Qwen3-0.6B config: 56 adapters injected, only q_proj/v_proj targeted, base weights frozen, adapter parameters trainable, and bf16 dtype preserved.\n- Broader collectable repository suite: 427 passed, 1 skipped, 1 unrelated environment failure in test_auto_model due missing modelscope/rapidfuzz and incompatible optional scientific packages on this host.\n- bash -n, compileall, and git diff --check passed.\n\nThe tests now cover target selection, initial equivalence, freezing/trainability, dtype, state-dict round-trip, train/eval transitions, and the no-match behavior. Thanks for closing the review gap.

PR #3459, opened 2026-08-03
test(nano): make LoRA tests deterministic

KamitobiHaru:
## Summary

Follow-up to #3456. While doing an adversarial re-verification of the merged tests, I found the state-dict round-trip test was **flaky**: it draws from the unseeded global RNG, and the merged vs unmerged fp32 forwards can differ by ~4e-5 (different op orderings), which exceeds `torch.allclose`'s default `rtol=1e-5`. In practice the test failed on roughly one in three full-file runs — i.e. random CI red builds. This PR makes the tests deterministic and tightens the honesty of the coverage.

## Changes

- **Seed the RNG per test** (`torch.manual_seed(0)` autouse fixture): every test is reproducible regardless of run order.
- **Compare merge/unmerge at a fp32-appropriate tolerance** (`rtol=1e-3`): fp32 non-associativity means the merged forward (`x @ (W+d).T`) and the unmerged forward (`x @ W'.T + (x @ A.T @ B.T) * scaling`) legitimately differ at ~1e-5. The tolerance is still discriminating — a genuinely broken merge differs by the full adapter magnitude (O(1–10)), which I verified adversarially (a "mark merged but don't fold the weight" bug gives `max_abs_diff ≈ 12`).
- **Add `test_mark_only_lora_as_trainable_keeps_adapters_trainable`**: the original reply described the `lora_only=true` flow but didn't actually test it. This test exercises the real `mark_only_lora_as_trainable` used by `train.py`/`train_ds.py`: after unfreezing the base model it re-freezes every non-LoRA parameter while `lora_A`/`lora_B` stay trainable.

## Verification

- 8/8 tests pass, stable across repeated runs; the full nano set is **35 passed** (27 existing + 8).
- On upstream `main` without the LoRA injection all 8 fail with `AttributeError` (tests genuinely exercise the new path).

### DIFF
--- a/tests/test_fun_asr_nano_lora_injection.py
+++ b/tests/test_fun_asr_nano_lora_injection.py
@@ -22,6 +22,7 @@ Coverage requested in review (modelscope/FunASR#3456):
 
 import os
 
+import pytest
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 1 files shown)


LauraGPT (approved):
Locally verified against the exact PR head. The complete nano suite passes (35/35), Ruff/compile/diff checks pass, and a 1,000-seed stress run reproduced 97 default-tolerance merge comparison failures plus 37 round-trip failures. With this PR's tolerance there were zero false failures, while an adversarial merge-without-weight-fold mutation was rejected in all 1,000 seeds (minimum max-absolute error 3.28 versus a valid maximum of 5.72e-6). The added trainability test also exercises the documented `lora_only` path.

PR #3471, opened 2026-08-05
fix(train): initialize DeepSpeed mode from top-level config

YuCeong-May:
## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [ ] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [ ] Docs or links checked

### DIFF
--- a/funasr/bin/train_ds.py
+++ b/funasr/bin/train_ds.py
@@ -83,9 +83,12 @@ def main(**kwargs):
     if local_rank == 0:
         tables.print()
 
-    use_ddp = world_size > 1
     use_fsdp = kwargs.get("use_fsdp", False)
     use_deepspeed = kwargs.get("use_deepspeed", False)
+    deepspeed_config = kwargs.get("deepspeed_config", "")
+    if use_deepspeed and use_fsdp:
+        raise ValueError("use_deepspeed and use_fsdp cannot be enabled at the same time")
+    use_ddp = world_size > 1 and not use_deepspeed and not use_fsdp
     if use_deepspeed:
         logging.info(f"use_deepspeed: {use_deepspeed}")
         deepspeed.init_distributed(dist_backend=kwargs.get("backend", "nccl"))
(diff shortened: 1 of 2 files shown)


LauraGPT (changes requested):
Thanks for fixing the DeepSpeed mode propagation. There is one blocking compatibility issue before this can merge.

All current DeepSpeed training examples put both settings under `train_conf`, for example:

```
++train_conf.use_deepspeed=false
++train_conf.deepspeed_config=...
```

This patch also passes `use_deepspeed=` and `deepspeed_config=` explicitly and then expands `**kwargs.get("train_conf")`. Python therefore raises before entering `Trainer`:

```
TypeError: Trainer() got multiple values for keyword argument use_deepspeed
```

Please normalize the effective values once, with top-level values taking precedence and `train_conf` retained as a backwards-compatible fallback. Then pass a copied `train_conf` with the consumed distributed-mode keys removed, so `use_ddp`, `use_fsdp`, `use_deepspeed`, and `deepspeed_config` cannot be supplied twice.

Please also add focused regression coverage for:
1. the existing nested `train_conf` form;
2. the new top-level form overriding nested defaults;
3. rejecting simultaneous FSDP and DeepSpeed.

`compileall` passes because this is a call-time configuration failure, so a behavioral test is needed.

LauraGPT (approved):
Re-reviewed head 241dea6352912336cf57d8d0be3f93533c19eb2e. The blocking duplicate-key path is resolved.

Verified:
- existing nested train_conf DeepSpeed settings remain supported
- top-level values override nested defaults
- consumed distributed keys are removed before Trainer receives **trainer_conf
- FSDP plus DeepSpeed is rejected
- DDP is disabled when either FSDP or DeepSpeed is selected
- focused tests: 3 passed
- additional precedence/invalid-combination matrix passed
- py_compile and diff checks passed
- synthetic merge with current main d410a565 is clean and limited to the intended two files

Approved.

PR #3472, opened 2026-08-05
fix(train): align DDP no_sync with gradient accumulation

YuCeong-May:
## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [ ] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [ ] Docs or links checked

### DIFF
--- a/funasr/train_utils/trainer_ds.py
+++ b/funasr/train_utils/trainer_ds.py
@@ -604,7 +604,9 @@ class Trainer:
 
             my_context = nullcontext
             if self.use_ddp or self.use_fsdp:
-                my_context = model.no_sync if batch_idx % accum_grad != 0 else my_context
+                my_context = (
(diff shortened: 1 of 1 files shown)


LauraGPT (approved):
Verified against current main and with a two-rank CPU/Gloo DDP reproduction.

The old batch_idx % accum_grad condition synchronizes the first microbatch and suppresses synchronization on the optimizer-step microbatch, leaving rank-local accumulated gradients (4.0 vs 7.0 in the reproduction). The proposed (batch_idx + 1) % accum_grad condition keeps early microbatches inside no_sync() and synchronizes the final microbatch, producing identical accumulated gradients on both ranks (5.5 vs 5.5).

Additional checks:
- synthetic merge with current main is clean and changes only this condition
- python -m py_compile funasr/train_utils/trainer_ds.py
- git diff --check

Approved.

PR #3480, opened 2026-08-08
llama.cpp: add srt output

gumblex:
## Summary
Add .srt format output for llama.cpp runtime

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Example or demo
- [x] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

- [ ] Docs or links checked

## User impact

Users want to use llama.cpp binary to directly generate usable subtitles.


LauraGPT (approved):
Built all three llama.cpp targets from the pinned standalone CMake project. Added maintainer fixes in 3ab58f9d to preserve Fun-ASR-Nano default output (`/sil` and trailing newline), corrected CLI usage, and documented `--srt`. Verified timestamp edge cases and complete SRT output byte-for-byte. Ready to merge.

PR #3494, opened 2026-08-14
feat(vad): add Silero VAD adapter

wuhongsheng:
## Summary
- add `vad_model="silero-vad"` backed by the official `silero-vad` Python package
- preserve the existing AutoModel VAD → ASR pipeline contract by returning millisecond segments
- add optional dependency, documentation, and focused tests

## Validation
- `python -m pytest -q tests/test_silero_vad_adapter.py`
- README example tested with `cloth_20220629_clean_01.wav` (non-empty transcript and 329 timestamps)

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,12 @@ from funasr import AutoModel
 model = AutoModel(model="paraformer-zh", vad_model="fsmn-vad", punc_model="ct-punc", spk_model="cam++", device="cuda")
 result = model.generate(input="https://isv-data.oss-cn-hangzhou.aliyuncs.com/ics/MaaS/ASR/test_audio/asr_example_zh.wav", hotword="关键词 20")
 
+# Optional Silero VAD (install first: python -m pip install "funasr[silero]")
+model = AutoModel(
+    model="paraformer-zh", vad_model="silero-vad", device="cuda",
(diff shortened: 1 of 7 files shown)


LauraGPT:
Maintainer validation completed on `e8ea59a02c9362ae78e6730ca574862f5aedd6c7`. I pushed two focused fixes before moving this out of draft:

- keep TorchScript input tensors on the adapter/model device so CUDA ASR + Silero VAD does not mix a CUDA model with CPU audio
- keep ONNX input tensors on CPU because the official wrapper calls `Tensor.numpy()`
- reject negative and sub-millisecond max segment values that would otherwise produce a zero-width split loop
- add the optional Silero path to the main English README

Validation:
- `python -m pytest -q tests/test_silero_vad_adapter.py` -> **7 passed** with `silero-vad==6.2.1`
- real 70.47 s WAV through the TorchScript alias -> **42 valid segments**
- the same WAV through `silero_onnx=True` with `onnxruntime==1.28.0` -> **42 valid segments**
- Black check, `py_compile`, and `git diff --check` passed
- repository `contract` check passed

LauraGPT (approved):
Reviewed the final maintainer-tested head. TorchScript and ONNX device behavior, segment bounds, optional dependency path, and documentation are covered.

PR #3518, opened 2026-08-20
fix: use torch.amp via compatibility shim to avoid deprecated torch.cuda.amp warnings

### LINKED ISSUE #3517: `torch.cuda.amp.autocast` / `GradScaler` usage across FunASR emits FutureWarning on torch ≥2.4 — migrate to `torch.amp`
## Summary

FunASR uses the deprecated `torch.cuda.amp` APIs in **37 files**. The
hot paths that actually call them:

| File | Line |
|---|---|
| `funasr/models/llm_asr/model.py` | 8 `from torch.cuda.amp import autocast`; 7 call sites (586, 621, 886, 1162, 1206, 1554 `with torch.cuda.amp.autocast(...)`) |
| `funasr/train_utils/trainer_ds.py` | 9 `from torch.cuda.amp import autocast, GradScaler`; 35 `with torch.cuda.amp.autocast(enabled=True, dtype=dtype, ...)` |
| `funasr/models/bat/model.py` | 27 `from torch.cuda.amp import autocast` (guard only checks torch < 1.6, not the deprecation) |
| …34 more files reference `torch.cuda.amp` | see below |

Verified on torch 2.11.0:

```
>>> with torch.cuda.amp.autocast():
FutureWarning: `torch.cuda.amp.autocast(args...)` is deprecated. Please
use `torch.amp.autocast('cuda', args...)` instead.
```

`autocast` has been deprecated since torch 2.4 and `GradScaler` since
torch 2.3. Both still work but print a `FutureWarning` on every call —
the current import guard in `bat/model.py` only covers torch < 1.6 and
does not protect against this.

## Suggested fix

The replacement APIs (`torch.amp.autocast('cuda', ...)` since 2.0,
`torch.amp.GradScaler('cuda', ...)` since 2.3) are available in all
current torch versions. If older torch versions must remain supported,
gate on the version — 2.3 is the boundary because the import below needs
both names, and torch < 2.3's `torch.cuda.amp` is not yet deprecated:

```python
if LooseVersion(torch.__version__) >= LooseVersion("2.3"):
    from torch.amp import autocast, GradScaler
else:
    from torch.cuda.amp import autocast, GradScaler
```

## Installation has no torch version constraint

FunASR is a single-package monorepo (one `setup.py` → `funasr`; `runtime/`
holds platform deployments). The install path does not bound torch:

- `setup.py` `install_requires` **does not list torch at all** — pip
  never installs or constrains it.
- [README.md:38-44](https://github.com/modelscope/FunASR/blob/main/README.md)
  instructs `pip install torch torchaudio` (latest PyPI = torch 2.11),
  then `pip install funasr`.

So on a fresh install with current PyPI, torch 2.11 is the documented
outcome and the deprecation warnings fire on every autocast/GradScaler
use.

(Also observed: the Triton GPU deployment image
[`runtime/triton_gpu/Dockerfile/Dockerfile.server`](https://github.com/modelscope/FunASR/blob/main/runtime/triton_gpu/Dockerfile/Dockerfile.server)
pins `pip3 install torch==2.4.1 torchaudio==2.4.1` — pinned to match the
prebuilt kaldifeat CUDA wheel. torch 2.4.1 is itself the first release
where `torch.cuda.amp.autocast` is deprecated, so even that pinned
deployment ships the deprecated API.)

## Files referencing torch.cuda.amp (37 total)

`funasr/bin/train.py`, `funasr/bin/train_ds.py`, `funasr/models/campplus/model.py`,
`funasr/models/data2vec/data2vec.py`, `funasr/models/llm_asr/model.py`,
`funasr/models/xvector/e2e_sv.py`, `funasr/models/sanm_kws/model.py`,
`funasr/models/fsmn_kws/model.py`, `funasr/models/e_paraformer/pif_predictor.py`,
`funasr/models/lcbnet/model.py`, `funasr/models/scama/model.py`,
`funasr/models/emotion2vec/model.py`, `funasr/train_utils/trainer.py`,
`funasr/models/uniasr/model.py`, `funasr/models/fsmn_kws_mt/model.py`,
`funasr/train_utils/trainer_ds.py`, `funasr/models/transducer/model.py`,
`funasr/models/mfcca/e2e_asr_mfcca.py`, `funasr/models/sa_asr/e2e_sa_asr.py`,
plus 18 more under `funasr/models/` and `funasr/train_utils/`.

xyf5432:
**Fixes #3517** — `torch.cuda.amp.autocast` has been deprecated since torch 2.4 and `torch.cuda.amp.GradScaler` since torch 2.3.

## Summary

Both APIs still work but emit a `FutureWarning` when used; the
replacement device-agnostic APIs live in `torch.amp`.

**Why migrate now, beyond warning noise:** two entry points
(`funasr/utils/export_utils.py`, `funasr/bin/realtime_ws.py`) already
call `warnings.filterwarnings("ignore")` at module level, so the
`FutureWarning` is not visible on those paths today. The real risk is the
announced removal of `torch.cuda.amp` ("will be removed in a future
release", no version pinned; still present in 2.11): the moment torch
drops it, all 37 `from torch.cuda.amp import ...` sites in this repo
raise `ImportError` at import time — a hard failure no warning filter
can mask.

Changes:
- Adds `funasr/utils/amp.py`, a compatibility shim that re-exports
  `torch.amp` names when available (torch >= 2.3) and falls back to
  `torch.cuda.amp` otherwise.
- The shim supplies `device_type="cuda"` by default for `autocast`,
  matching the old `torch.cuda.amp.autocast` signature — existing call
  sites (`with autocast(enabled=..., dtype=...)`) work unchanged.
- Replaces all 37 `from torch.cuda.amp import ...` imports with the shim.
- Rewrites the 7 `torch.cuda.amp.autocast(...)` attribute call sites
  (llm_asr/model.py ×6, trainer_ds.py ×1) to use the imported name, so
  the deprecation path is actually avoided.

## Type of change

- [√] Bug fix — deprecated API usage that becomes a hard `ImportError` when torch removes `torch.cuda.amp`
- [ ] Documentation
- [ ] Example or demo
- [ ] Runtime or deployment
- [ ] Benchmark or evaluation
- [ ] Model/training change

## Validation

Verified on torch 2.11.0 with `warnings.simplefilter("error", FutureWarning)`
(re-asserted after importing funasr — `export_utils.py` / `realtime_ws.py`
set a catch-all `filterwarnings("ignore")` at module level that would
otherwise mask the test filter):

- `from funasr.utils.amp import autocast, GradScaler` — no FutureWarning
- `with autocast():` / `with autocast(enabled=True, dtype=None, cache_enabled=False):` — no FutureWarning
- `GradScaler(enabled=True)` — no FutureWarning
- explicit `device_type` and positional args still pass through
- torch < 2.3 fallback path is identical to today (not yet deprecated there)

- [√] `python -m compileall funasr examples tests` — passes. (The SyntaxWarnings in `funasr/models/fsmn_kws/encoder.py` are pre-existing invalid escape sequences; that file is not touched by this PR.)
- [ ] Docs or links checked

## User impact

Users of FunASR on torch >= 2.3 — including the current PyPI default
torch 2.11, which the documented README install path resolves to — no
longer see the deprecation `FutureWarning` on training/inference paths,
and every entry point keeps importing when torch eventually removes
`torch.cuda.amp`. No behavior change on torch < 2.3.

## Notes for reviewers

- The shim mirrors the old `torch.cuda.amp.autocast` signature via a
  `device_type="cuda"` default, because `torch.amp.autocast` requires a
  positional `device_type`.
- Migration boundary is 2.3, not 2.0: the shim imports both `autocast`
  and `GradScaler` from `torch.amp` together, and `GradScaler` only
  exists there since 2.3.
- `bat/model.py` keeps its torch < 1.6 guard — the shim's fallback
  imports `torch.cuda.amp`, which only exists since 1.6.
- The two entry points that already silence all warnings see no visible
  change; the fix is future-proofing for the removal.

## Reference

- [PyTorch AMP docs](https://pytorch.org/docs/stable/amp.html) — official deprecation notice (`torch.cuda.amp.autocast` / `GradScaler` deprecated since 2.4 / 2.3, "will be removed in a future release")

### DIFF
--- a/funasr/bin/train.py
+++ b/funasr/bin/train.py
@@ -16,7 +16,7 @@ from contextlib import nullcontext
 import torch.distributed as dist
 
 from omegaconf import DictConfig, OmegaConf
-from torch.cuda.amp import autocast, GradScaler
+from funasr.utils.amp import autocast, GradScaler
 from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.algorithms.join import Join
--- a/funasr/bin/train_ds.py
+++ b/funasr/bin/train_ds.py
@@ -15,7 +15,7 @@ from contextlib import nullcontext
 import torch.distributed as dist
 
 from omegaconf import DictConfig, OmegaConf
-from torch.cuda.amp import autocast, GradScaler
+from funasr.utils.amp import autocast, GradScaler
 from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.algorithms.join import Join
--- a/funasr/models/bat/model.py
+++ b/funasr/models/bat/model.py
@@ -24,7 +24,7 @@ from funasr.models.transducer.beam_search_transducer import BeamSearchTransducer
 
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/bicif_paraformer/model.py
+++ b/funasr/models/bicif_paraformer/model.py
@@ -26,7 +26,7 @@ from funasr.utils.load_utils import load_audio_text_image_video, extract_fbank
 from funasr.train_utils.device_funcs import to_device
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/campplus/model.py
+++ b/funasr/models/campplus/model.py
@@ -26,7 +26,7 @@ from funasr.models.campplus.components import (
 
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/contextual_paraformer/model.py
+++ b/funasr/models/contextual_paraformer/model.py
@@ -29,7 +29,7 @@ from funasr.utils.load_utils import load_audio_text_image_video, extract_fbank
 
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/ct_transformer/model.py
+++ b/funasr/models/ct_transformer/model.py
@@ -23,7 +23,7 @@ try:
 except:
     pass
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/ct_transformer_streaming/model.py
+++ b/funasr/models/ct_transformer_streaming/model.py
@@ -16,7 +16,7 @@ from funasr.models.ct_transformer.utils import split_to_mini_sentence, split_wor
 
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
-    from torch.cuda.amp import autocast
+    from funasr.utils.amp import autocast
 else:
     # Nothing to do if torch<1.6.0
     @contextmanager
--- a/funasr/models/data2vec/data2vec.py
+++ b/funasr/models/data2vec/data2vec.py
@@ -22,7 +22,7 @@ from funasr.frontends.abs_frontend import AbsFrontend
 from funasr.train_utils.device_funcs import force_gatherable
 
 if LooseVersion(torch.__version__) >= LooseVersion("1.6.0"):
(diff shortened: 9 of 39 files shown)


LauraGPT:
Pushed signed commit `a3009e31` to preserve the legacy `autocast(enabled, dtype, cache_enabled)` positional mapping. The prior head sent `autocast(False)` as `device_type=False`; the regression fails there and passes now.

Verification:
- related tests: 18/18 passed
- remaining collectable suite: 466 passed, 8 skipped, 1 deselected
- `compileall` and diff checks passed

The deselected `AutoModel` case is an existing environment failure caused by missing `modelscope`/registration dependencies. Full collection also has eight unrelated dependency errors, including the local NumPy/SciPy ABI mismatch.

LauraGPT (changes requested):
The fallback cannot run on PyTorch releases that do not expose `torch.amp`. PyTorch v1.9.1 provides `torch.cuda.amp` but has no `torch.amp` package: https://github.com/pytorch/pytorch/tree/v1.9.1/torch/cuda/amp

The condition accesses `torch.amp` before `hasattr`, so importing the shim raises `AttributeError` before reaching the `else` branch. I reproduced this with a torch stub that provides `torch.cuda.amp.autocast` and `GradScaler` but no `torch.amp`.

Please retrieve `torch.amp` safely (for example, with `getattr(torch, "amp", None)`) and add a regression test for the no-`torch.amp` path. Existing model guards retain PyTorch 1.6+ behavior, so this is within the compatibility scope of this PR.

LauraGPT (approved):
Verified exact head `1533c473`: the old-PyTorch import regression fails on the previous head with `AttributeError: module torch has no attribute amp` and passes after the defensive namespace lookup. Focused AMP compatibility tests pass 3/3; compileall, diff checks, SSH signature, and DCO are clean. This resolves the requested compatibility blocker without changing the modern torch.amp path.

PR #3524, opened 2026-08-22
Raise real exceptions instead of strings in the runtime Python bindings

### LINKED ISSUE #3523: funasr_onnx: bare `raise "string"` masks the real exception (TypeError: exceptions must derive from BaseException)
Several error paths in `funasr_onnx` do `raise "some message"`. That is invalid in Python 3 — a string is not an exception — so the interpreter discards the intended message and raises `TypeError: exceptions must derive from BaseException` instead.

Combined with the surrounding bare `except:`, the actual cause is swallowed entirely.

### Where

`runtime/python/onnxruntime/funasr_onnx/sensevoice_bin.py` (current `main`), lines 50, 54, 66:

```python
try:
    from funasr import AutoModel
except:
    raise "You are exporting onnx, please install funasr and try it again. ..."
```

The same pattern appears in `paraformer_online_bin.py` and other `*_bin.py` files.

### What the user sees

```
File ".../funasr_onnx/sensevoice_bin.py", line 66, in __init__
    raise "You are exporting onnx, please install funasr and try it again..."
TypeError: exceptions must derive from BaseException
```

### What was actually wrong

In my case `funasr` **was** installed. The import failed on a missing transitive
dependency:

```
File ".../funasr/utils/load_utils.py", line 9, in <module>
    import torchaudio
ModuleNotFoundError: No module named 'torchaudio'
```

The message told me to install a package I already had, while hiding the one I was
missing. Diagnosing it required reading the library source and reproducing the import
by hand.

I hit the same thing a second time on `onnxscript` (required by newer `torch.onnx`),
and again when loading a model from a local path — line 54 fires whenever the path
check fails for *any* reason, so a simple wrong-path mistake also surfaces as
`TypeError`.

### Suggested fix

```python
try:
    from funasr import AutoModel
except ImportError as e:
    raise ImportError(
        "Exporting ONNX requires funasr: pip3 install -U funasr"
    ) from e
```

`raise ... from e` keeps the original traceback, so a missing `torchaudio` stays
visible. Narrowing `except:` to `except ImportError:` also stops unrelated failures
from being reported as a missing package.

Happy to send a PR if that would help.

LeonardNJU:
Fixes #3523.

`raise "some message"` is invalid in Python 3 — a `str` is not an exception. The
interpreter discards the intended message and raises

```
TypeError: exceptions must derive from BaseException
```

instead. Paired with the surrounding bare `except:`, the original cause is lost as well,
so the user is told to install a package they may already have while the real failure is
invisible.

### What this looked like in practice

`funasr` **was** installed. The import failed on a missing transitive dependency:

```
File ".../funasr/utils/load_utils.py", line 9, in <module>
    import torchaudio
ModuleNotFoundError: No module named 'torchaudio'
```

What I saw:

```
File ".../funasr_onnx/sensevoice_bin.py", line 66, in __init__
    raise "You are exporting onnx, please install funasr and try it again..."
TypeError: exceptions must derive from BaseException
```

The message pointed at the wrong package, and the traceback pointed at the wrong error.
Diagnosing it meant reading the library source and reproducing the import by hand. I hit
the same wall a second time on `onnxscript` (required by newer `torch.onnx`), and a third
time when loading from a local path — the `model_dir` branch fires on *any* failure, so a
mistyped path also surfaced as `TypeError`.

### The change

All 30 occurrences under `runtime/python` (7 files, both `funasr_onnx` and
`funasr_torch`):

```python
# before
try:
    from funasr import AutoModel
except:
    raise "You are exporting onnx, please install funasr and try it again. ..."

# after
try:
    from funasr import AutoModel
except ImportError as e:
    raise ImportError(
        "You are exporting onnx, please install funasr and try it again. ..."
    ) from e
```

- guarded imports → `except ImportError as e:` / `raise ImportError(...) from e`
- everything else → `except Exception as e:` / `raise RuntimeError(...) from e`

`from e` preserves the original traceback, which is the point — a missing `torchaudio`
now stays visible under the friendly message. Narrowing the bare `except:` also stops
unrelated failures from being reported as a missing package.

**Messages are unchanged.** The only behavioural difference is the exception type, which
previously could not be caught by anything anyway.

### Verification

- `raise "` no longer appears anywhere under `runtime/python`
- every changed file passes `python -m py_compile`
- chaining behaves as intended:

```
外层信息: please install funasr: pip3 install -U funasr
真实原因: No module named 'nonexistent_module_xyz'
```

I deliberately did **not** run `black` over these files. It reformats unrelated code in
them (the tree is not currently black-clean), which would have buried a 140-line fix in a
280-line diff. The new code follows the project's 100-column style by hand. Happy to run
the formatter if you would rather have it.

Two occurrences outside `runtime/python` are left alone as out of scope:
`funasr/models/qwen3_asr/model.py` and `funasr/download/runtime_sdk_download_tool.py`.

Context: found while building a local-ASR adapter for Jitsi Meet
(https://github.com/LeonardNJU/jitsi-local-asr) that runs SenseVoice and streaming
Paraformer via `funasr_onnx` on an ARM box.

### DIFF
--- a/runtime/python/libtorch/funasr_torch/paraformer_bin.py
+++ b/runtime/python/libtorch/funasr_torch/paraformer_bin.py
@@ -37,14 +37,18 @@ class Paraformer:
         if not Path(model_dir).exists():
             try:
                 from modelscope.hub.snapshot_download import snapshot_download
-            except:
-                raise "You are exporting model from modelscope, please install modelscope and try it again. To install modelscope, you could:\n" "\npip3 install -U modelscope\n" "For the users in China, you could install with the command:\n" "\npip3 install -U modelscope -i https://mirror.sjtu.edu.cn/pypi/web/simple"
+            except ImportError as e:
+                raise ImportError(
+                    "You are exporting model from modelscope, please install modelscope and try it again. To install modelscope, you could:\n" "\npip3 install -U modelscope\n" "For the users in China, you could install with the command:\n" "\npip3 install -U modelscope -i https://mirror.sjtu.edu.cn/pypi/web/simple"
+                ) from e
             try:
                 model_dir = snapshot_download(model_dir, cache_dir=cache_dir)
-            except:
-                raise "model_dir must be model_name in modelscope or local path downloaded from modelscope, but is {}".format(
-                    model_dir
-                )
+            except Exception as e:
+                raise RuntimeError(
+                    "model_dir must be model_name in modelscope or local path downloaded from modelscope, but is {}".format(
+                        model_dir
+                    )
+                ) from e
 
         model_file = os.path.join(model_dir, "model.torchscript")
         if quantize:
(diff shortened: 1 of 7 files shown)


LauraGPT (approved):
Verified exact head `4999dc7632f31a353674af4c120ce36a3e1a25f2` locally:

- all seven changed Python files compile;
- `git diff --check` passes;
- an AST audit covers all 30 changed exception paths: 20 chained `ImportError` handlers, 10 chained `RuntimeError` handlers, and no string raises remain in the changed files.

The narrowed handlers preserve the original cause and no longer turn a useful error into `TypeError`. Approving.

PR #3526, opened 2026-08-26
Remove hard torchaudio dependency for inference; add kaldi-native-fbank fbank backend

li-lizhe:
Remove hard torchaudio dependency for inference; add kaldi-native-fbank fbank backend

Make torchaudio an optional dependency for the inference path so FunASR can run
on Ascend NPU (aarch64) servers, where no matching torchaudio wheel exists
(Huawei's Ascend torch is version-decoupled from the PyTorch upstream release
line). This is the root cause behind the confusing missing-dependency import
failures reported in #3523.

## Changes

- Add `funasr/utils/fbank.py`: an fbank shim that delegates to
  `torchaudio.compliance.kaldi` when torchaudio is installed, otherwise falls
  back to kaldi-native-fbank (a torch-free C++ library with aarch64 wheels).
- Switch `funasr/frontends/wav_frontend.py`, `funasr/utils/speaker_utils.py`
  and `funasr/models/campplus/utils.py` to the shim.
- Make torchaudio optional in the remaining inference/training modules
  (paraformer_v2, fun_asr_nano, dataset preprocessors) and add a shared guard
  (`funasr/utils/torchaudio_compat.py`) so operations that still need torchaudio
  fail with an actionable error instead of a bare AttributeError or a silently
  wrong result.

No device-specific branching is introduced — a pure backend fallback that also
benefits CPU-only and other accelerator platforms.

## Changes since first review

- Declare kaldi-native-fbank as an optional `knf` extra in `setup.py`; the fbank
  shim now raises an actionable ImportError only when `fbank()` is called with
  neither backend installed (previously import itself could fail with a bare
  ModuleNotFoundError).
- Harden the fallback contract: handle `channel` explicitly and reject
  torchaudio-only options (`subtract_mean`, `min_duration`, VTLN warping,
  non-default `blackman_coeff`) with NotImplementedError instead of silently
  dropping them via `**kwargs`. Multi-channel waveforms without an explicit
  `channel` are rejected (kaldi-native-fbank is single-channel).
- `forced_align` now propagates the missing-torchaudio error instead of
  returning an empty alignment; paraformer_v2 `force_align` and speed
  perturbation guard torchaudio the same way.
- Add `tests/test_torchaudio_optional.py`: regression tests that import the
  affected modules with torchaudio absent and exercise the fallback contract.

This PR now includes what was previously proposed separately in #3527 (optional
torchaudio in paraformer_v2 / fun_asr_nano / dataset preprocessors) so the
change is atomic.

## Verification

- `tests/test_torchaudio_optional.py` passes in an environment with torchaudio
  absent and kaldi-native-fbank present.
- End-to-end on Ascend 910B (npu:0) with torchaudio absent: paraformer-large
  loads and infers correctly — 70.47s of audio in 1.15s, RTF ~0.016 (~60x
  realtime), fbank extracted via the kaldi-native-fbank fallback.

Related: #3523

### DIFF
--- a/funasr/datasets/audio_datasets/preprocessor.py
+++ b/funasr/datasets/audio_datasets/preprocessor.py
@@ -7,7 +7,10 @@ import librosa
 import torch.distributed as dist
 from typing import Collection
 import torch
-import torchaudio
+try:
+    import torchaudio
+except ImportError:
+    torchaudio = None
 from torch import nn
 import random
 import re
--- a/funasr/datasets/audio_datasets/preprocessor.py
+++ b/funasr/datasets/audio_datasets/preprocessor.py
@@ -16,6 +16,7 @@ import random
 import re
 from funasr.tokenizer.cleaner import TextCleaner
 from funasr.register import tables
+from funasr.utils.torchaudio_compat import require_torchaudio
 
 
 @tables.register("preprocessor_classes", "SpeechPreprocessSpeedPerturb")
@@ -42,6 +43,7 @@ class SpeechPreprocessSpeedPerturb(nn.Module):
             return waveform
         speed = random.choice(self.speed_perturb)
         if speed != 1.0:
+            torchaudio = require_torchaudio("speed perturbation")
             if not isinstance(waveform, torch.Tensor):
                 waveform = torch.tensor(waveform)
             waveform, _ = torchaudio.sox_effects.apply_effects_tensor(
--- a/funasr/datasets/llm_datasets/preprocessor.py
+++ b/funasr/datasets/llm_datasets/preprocessor.py
@@ -7,7 +7,10 @@ import librosa
 import torch.distributed as dist
 from typing import Collection
 import torch
-import torchaudio
+try:
+    import torchaudio
+except ImportError:
+    torchaudio = None
 from torch import nn
 import random
 import re
--- a/funasr/frontends/wav_frontend.py
+++ b/funasr/frontends/wav_frontend.py
@@ -5,7 +5,7 @@ import copy
 import numpy as np
 import torch
 import torch.nn as nn
-import torchaudio.compliance.kaldi as kaldi
+from funasr.utils import fbank as kaldi
 from torch.nn.utils.rnn import pad_sequence
 
 import funasr.frontends.eend_ola_feature as eend_ola_feature
--- a/funasr/models/campplus/utils.py
+++ b/funasr/models/campplus/utils.py
@@ -16,7 +16,7 @@ from typing import Union
 from pathlib import Path
 from typing import Generator, Union
 from abc import ABCMeta, abstractmethod
-import torchaudio.compliance.kaldi as Kaldi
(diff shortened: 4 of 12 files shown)


LauraGPT (changes requested):
Two blocking contract gaps remain before this can replace torchaudio on the target installation path.

1. `funasr/utils/fbank.py` imports `kaldi_native_fbank` immediately when torchaudio is absent, but `kaldi-native-fbank` is not declared in `install_requires` or an optional extra. A clean environment without torchaudio therefore still fails at import with `ModuleNotFoundError`. Please declare an installable dependency path (base dependency or documented extra), provide a clear missing-dependency error, and test the torchaudio-absent path.

2. The shim accepts arbitrary `**kwargs` and silently ignores unsupported torchaudio parameters. For example, torchaudio supports a 2-D waveform with `channel=0`; the fallback drops `channel` and then fails while squeezing. Please either map the supported contract or reject unsupported options explicitly and narrow the drop-in claim.

The focused existing tests pass, but they do not exercise an environment where torchaudio is completely absent. Please also link #3523 in the PR body. #3527 is functionally dependent on this fallback and should be rebased after this contract is settled.

li-lizhe:
@LauraGPT Thanks for the thorough review. Both sets of feedback are addressed in the latest push, and #3527 has been folded into this PR so the change is atomic:

1. kaldi-native-fbank is now declared as an optional `knf` extra in setup.py; the fbank shim only raises an actionable ImportError when fbank() is called with neither backend available (import itself no longer fails).
2. The fallback handles `channel` explicitly — selects a channel from 2-D waveforms, rejects multi-channel without an explicit channel — and rejects torchaudio-only options (subtract_mean, min_duration, VTLN, blackman_coeff) with NotImplementedError instead of silently dropping them.
3. Added a shared guard (funasr/utils/torchaudio_compat.py): forced alignment now propagates the missing-torchaudio error instead of returning []; paraformer_v2 force_align and speed perturbation guard torchaudio the same way.
4. Added tests/test_torchaudio_optional.py — regression tests that import the affected modules with torchaudio absent and exercise the fallback contract.
5. Linked #3523.

Re-verified end-to-end on Ascend 910B with torchaudio absent (70.47s audio, RTF ~0.016, fbank via kaldi-native-fbank fallback). Please take another look when you have a chance.

LauraGPT (approved):
The latest head addresses both blocking contracts. The `knf` extra and lazy error provide a clean install path, unsupported torchaudio options are rejected explicitly, channel selection is handled, and missing-torchaudio errors now propagate from the operations that still require it.

Fresh maintainer verification on exact head `7022ceb8`:

- optional-torchaudio plus audio-loading regressions: 24/24 passed
- real no-torchaudio `kaldi-native-fbank 1.22.3` WavFrontend smoke: 2/2 samples, finite 80-bin features
- torchaudio vs KNF parity: identical frame lengths, max absolute feature delta 1.46e-4
- focused compileall and `git diff --check`: passed

Approved.

PR #3527, opened 2026-08-26
Make torchaudio import optional in paraformer_v2, fun_asr_nano and dataset preprocessors

li-lizhe:
Make torchaudio import optional in paraformer_v2, fun_asr_nano and dataset preprocessors

Follow-up to the inference-path change: make the remaining top-level
torchaudio imports optional so these modules import cleanly when
torchaudio is not installed (e.g. on Ascend NPU servers).

- funasr/models/paraformer_v2_community/model.py
- funasr/models/fun_asr_nano/tools/utils.py
- funasr/datasets/audio_datasets/preprocessor.py
- funasr/datasets/llm_datasets/preprocessor.py

Functions that still require torchaudio (forced_align, Resample) degrade
to a clear error only when called, rather than a hard import failure.

### DIFF
--- a/funasr/datasets/audio_datasets/preprocessor.py
+++ b/funasr/datasets/audio_datasets/preprocessor.py
@@ -7,7 +7,10 @@ import librosa
 import torch.distributed as dist
 from typing import Collection
 import torch
-import torchaudio
+try:
+    import torchaudio
+except ImportError:
+    torchaudio = None
 from torch import nn
 import random
 import re
--- a/funasr/datasets/llm_datasets/preprocessor.py
+++ b/funasr/datasets/llm_datasets/preprocessor.py
@@ -7,7 +7,10 @@ import librosa
 import torch.distributed as dist
 from typing import Collection
 import torch
-import torchaudio
+try:
+    import torchaudio
+except ImportError:
+    torchaudio = None
 from torch import nn
 import random
 import re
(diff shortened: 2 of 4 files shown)


LauraGPT (changes requested):
Making imports optional is useful, but the promised call-time behavior is not implemented yet.

When torchaudio is absent, `load_audio()` still dereferences `torchaudio.transforms` and raises `AttributeError: NoneType has no attribute transforms`; Paraformer resampling and speed perturb have the same pattern. `forced_align()` is more dangerous because its broad exception handler swallows the missing dependency and returns an empty alignment, turning a missing runtime requirement into a silent wrong result.

Please add one shared, explicit dependency guard with actionable installation guidance at every operation that still requires torchaudio, add regressions that load these modules with torchaudio absent, and ensure forced alignment propagates that dependency error instead of returning `[]`. This PR is also functionally dependent on #3526: by itself, other core inference imports still fail through the fbank path, so please rebase it after #3526 is corrected or combine the changes atomically. Please link #3523 in the PR body.

li-lizhe:
Superseded by #3526. The optional-torchaudio changes for paraformer_v2 / fun_asr_nano / dataset preprocessors have been folded into #3526 so the change is atomic (per review feedback that this PR was functionally dependent on the fbank fallback and should be combined or rebased). Closing this in favor of #3526.

PR #3540, opened 2026-08-28
fix: guard against IndexError on an empty segment in timestamp_sentence_en

AmirF194:
`timestamp_sentence_en` indexes `sentence_text[0]` and `sentence_text_seg[-1]` without checking either string is non-empty. When a punctuation/timestamp entry has no matching word (for example a whitespace-only text segment, where `text_postprocessed.split()` yields fewer items than `punc_id_list`), both strings stay empty on that iteration and the unguarded indexing raises `IndexError: string index out of range`, crashing any `en_post_proc` timestamp generation that hits it.

The sibling function `timestamp_sentence` already guards the same pattern (`sentence_text_seg and sentence_text_seg[-1] == " "`, added in #2429 for the identical class of crash). `timestamp_sentence_en` never got the equivalent guard on either of its two indexing sites and has carried this since it was introduced.

Fix mirrors #2429's guard shape at both sites in `timestamp_sentence_en`.

Verified in a clean container against this branch:
- Before the fix, calling the real `timestamp_sentence_en` with a whitespace-only segment (`punc_id_list=[3]`, `timestamp_postprocessed=[[0, 100]]`, `text_postprocessed=" "`) raises `IndexError: string index out of range`. After the fix, it returns normally.
- A normal multi-word English sentence with punctuation (single and multi-sentence cases, plus the `return_raw_text=True` path) splits and timestamps identically before and after the change.
- `python -m compileall funasr examples tests` passes, matching this repo's own Python-only PR checklist item.

I did not add a new test file since none currently exists for `timestamp_tools.py` and #2429 also shipped without one; happy to add one if that is useful here.

### DIFF
--- a/funasr/utils/timestamp_tools.py
+++ b/funasr/utils/timestamp_tools.py
@@ -281,16 +281,16 @@ def timestamp_sentence_en(
 
         punc_id = int(punc_id) if punc_id is not None else 1
         sentence_end = timestamp[1] if timestamp is not None else sentence_end
-        sentence_text = sentence_text[1:] if sentence_text[0] == ' ' else sentence_text
+        if sentence_text.startswith(" "):
+            sentence_text = sentence_text[1:]
         if is_sentence_start:
             sentence_start = timestamp[0] if timestamp is not None else sentence_start
             is_sentence_start = False
         if punc_id > 1:
             is_sentence_start = True
             sentence_text += punc_list[punc_id - 2]
-            sentence_text_seg = (
-                sentence_text_seg[:-1] if sentence_text_seg[-1] == " " else sentence_text_seg
-            )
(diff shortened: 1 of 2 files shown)


LauraGPT (approved):
Validated and completed on exact head `0a90cb5ca1b18ed3ab732d3b06d14b5b5952a6d5`. The whitespace-only regression reproduces the original IndexError on the base and passes with the guarded `startswith`/`endswith` fix. Added `tests/test_timestamp_tools.py`; focused timestamp and punctuation tests pass 4/4, compileall and diff checks pass. The commit is GitHub Verified, DCO-signed, and credits Amir Fathi as co-author.

AmirF194:
Thanks for validating on the exact head and merging. Appreciate the quick review.

PR #3550, opened 2026-08-29
fix(runtime): decode SenseVoice F16 query embeddings

LauraGPT:
## Summary

- decode SenseVoice query embeddings according to their GGML tensor type instead of reinterpreting F16 storage as `float*`
- validate the embedding width and query-token bounds before inference
- update the SenseVoice numerical golden to the corrected ITN output

## Root cause

The published F16 GGUF stores `embed.weight` as F16 `[560, 16]`, while the runtime read it as F32. Query-token copies therefore read twice the actual storage width and crossed tensor bounds. Depending on process memory contents, CTC intermittently collapsed to blank tokens and returned an empty transcript with exit code 0.

Related adopter report: https://github.com/huaxin0/FunASR-GGML/issues/4

## Validation

- reproduced with the exact `runtime-llamacpp-v0.2.3` Linux AVX2 asset: 22 newline-only outputs among 298 completed sequential runs
- red/green source contract test for F16/F32 embedding decoding
- 100/100 fixed F16 runs: zero empty outputs, one output hash, all compute statuses successful
- corrected F16 output matches the Q8 model F32-embedding output byte-for-byte
- `python3 -m pytest -q runtime/llama.cpp/tests/*.py`: 22 passed
- `run_regression.sh` with the F16 model: SenseVoice passed
- Release build of `llama-funasr-sensevoice` succeeded


LauraGPT:
Exact-head release matrix completed successfully on `178a36718cb1110255296bdf4986ff4ebf295019`:

- 9/9 build jobs succeeded
- Linux x64, AVX2, Vulkan, ARM64
- macOS ARM64
- Windows x64, AVX2, Vulkan, CUDA

Evidence: https://github.com/modelscope/FunASR/actions/runs/33229633488

PR #3569, opened 2026-08-29
docs: expose MOSS LocalAI edge path

LauraGPT:
## Summary
- expose the third-party LocalAI / moss-transcribe.cpp GGUF path on the MOSS deployment page and selector
- distinguish the FunASR AutoModel contract from the independent C++ reimplementation
- add pinned upstream evidence and CPU / desktop-edge discovery without claiming unverified Windows support

## Verification
- python -m pytest -q web-pages/product-site/tests (95 passed)
- product-site build (29 pages)
- output validator (108 pages)
- generated Chinese and English MOSS pages contain the LocalAI evidence and interface markers
- git diff --check

### DIFF
--- a/web-pages/product-site/data/deployments.json
+++ b/web-pages/product-site/data/deployments.json
@@ -8,11 +8,11 @@
       "maturity": "community-verified",
       "selector_rank": 12,
       "workloads": ["batch", "private-api"],
-      "hardware": ["nvidia-gpu", "kubernetes"],
+      "hardware": ["cpu", "nvidia-gpu", "desktop-edge-gpu", "kubernetes"],
       "priorities": ["throughput"],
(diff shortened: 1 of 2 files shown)


LauraGPT:
Merged as 8b55e45334d8a55f6cb9522674e52a20ddfb2ab4 and deployed from that exact source.

Post-deploy verification:
- Chinese: https://www.funasr.com/deploy/moss-transcribe-diarize.html
- English: https://www.funasr.com/en/deploy/moss-transcribe-diarize.html
- both routes return HTTP 200 and expose LocalAI / moss-transcribe.cpp as a separate third-party C++/GGUF edge path
- both retain the OpenMOSS ownership, independent-output-contract, pinned-revision, and target-hardware validation boundaries

The first release attempt automatically restored the previous site because this host has no populated nginx PID file. The successful release used the verified nginx master PID, reran 95 site tests and the 108-page validator, and saved a rollback backup before switching the symlink.

Issue #914, opened 2023-09-02
VAD give wrong timestamp.


reopjk:
My audio length is 26s, however vad gives timestamp over 26s. Just like shown in the image.
![image](https://github.com/alibaba-damo-academy/FunASR/assets/61106156/ba368463-c02d-4c87-98b7-ea1c8af8f5f7)

PR #3579, opened 2026-08-30
feat(moss): add SGLang AutoModel backend

LauraGPT:
## Summary

- add backend="sglang" for the third-party OpenMOSS MOSS-Transcribe-Diarize model through the OpenAI-compatible transcription endpoint
- request verbose_json with max_new_tokens and normalize validated speaker segments into FunASR text, raw_text, timestamp, and sentence_info
- reject malformed labels, decreasing start times, and SGLang fallback S01 labels that are not backed by the raw model transcript
- preserve legal overlapping speech and SGLang timestamp clamping
- update bilingual guides, four deployment matrices, and the product deployment registry with the exact SGLang source pin

The model and weights remain published and maintained by OpenMOSS under Apache-2.0. FunASR provides only the ecosystem adapter and deployment contract. External vad_model and spk_model remain intentionally unsupported for this end-to-end model.

## Validation

- exact signed+DCO head: 783c24adfb7b8f327602d5c05af9b098bee1fbae
- 40 model, documentation, and product-registry tests passed
- MOSS product-page render test passed
- Black, JSON validation, compileall, and git diff check passed
- independent review found no remaining Critical or Important issues

## Runtime evidence boundary

No live SGLang worker was claimed in this change. The available H100 host has a CUDA 12.8 driver while the pinned current SGLang Omni path requires CUDA 13. The adapter is contract-tested against the exact public API/source behavior at SGLang Omni 3f819f9cdae3d4eeec22f73306c9067a1ec2542e. The published #914 H100 benchmark remains upstream runtime evidence only, not a FunASR adapter benchmark.


LauraGPT:
CI follow-up: the first contract run failed only its Black check. I reproduced the exact formatter locally with Black 25.1.0, confirmed the diff was two mechanical line wraps, amended the signed+DCO commit, and reran Black plus all 40 focused tests and the product-page render test before pushing exact head 783c24adfb7b8f327602d5c05af9b098bee1fbae.

PR #3589, opened 2026-08-31
fix(realtime): preserve aligned long-segment partials

LauraGPT:
## Summary

- preserve a proven long-segment partial when the completed decode collapses to one character
- tolerate small early insertions/corrections when detecting a later tail regression
- retain the existing long-segment, same-boundary, repeated-observation, and near-tail safety gates

## Root cause

The reconciliation path required a final transcript to contain at least two normalized characters and compared the final result with the best observed partial using a strict common prefix. The reporter's latest 1.4.11 traces expose both gaps:

- a complete partial was followed by a one-character final (`哎`), which the minimum-length check rejected
- a complete partial and truncated final differed by a small early insertion/correction, so the strict prefix comparison never reached the missing tail

The updated check uses ordered matching blocks. Catastrophically short finals must still match at least half of their normalized characters, while tail regressions require at least 75% final coverage and an aligned start.

## Verification

Exact signed+DCO head: `9aec3eeb67263bffada3e9fcf6863fb909867fef`

- focused realtime / postprocess / file-finalization suites: 122 passed, 1 environment skip
- Python compilation and `git diff --check`: passed
- H100, complete Nano checkpoint, speaker path enabled, reporter's 14.432-second `test_audio.wav`: 8/8 final responses retained the `胡总...长官` tail; 0/8 collapsed to one character
- full WebSocket traces, server trigger logs, SHA-256 manifest, and rollback patch are saved in the publication backup

Related to #3302. The issue should remain open until the reporter retests the original deployment; merging this PR is not reporter confirmation.

### DIFF
--- a/funasr/bin/realtime_ws.py
+++ b/funasr/bin/realtime_ws.py
@@ -13,6 +13,7 @@ Features:
 import asyncio
 from collections import deque
 import copy
+from difflib import SequenceMatcher
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


pengzhendong (approved):
LGTM

PR #3591, opened 2026-08-31
fix(realtime): accept bounded partial VAD overrun

LauraGPT:
## Summary

- accept a proven long-segment partial when its decode endpoint is slightly later than the VAD endpoint
- bound the allowed overrun to one realtime decode chunk
- add exact regression coverage for the reporter's `114430/126976/126740ms` trace and a farther-overrun rejection test

## Root cause

The follow-up on #3302 used exact merge `2b6294d69e6588a63c0ce06300f1b55ded71ae58` and still locked a complete partial down to `哎`. The last partial covered `114430-126976ms`, while FSMN-VAD emitted `114430-126740ms`.

The long-segment reconciliation gate only accepted `tail_gap_ms >= 0`. Normal decode/VAD scheduling therefore rejected this candidate solely because it ended 236ms after the VAD boundary, before the single-character regression check could run.

The new lower bound allows at most one decode chunk of overrun. Existing exact start alignment, repeated observations, long-segment duration, text alignment, and hallucination gates remain unchanged. A partial more than two chunks past this concrete boundary remains rejected by regression coverage.

## Verification

Exact signed+DCO head: `980765099dea8d9f34ea2f10889b3ab1c3876173`

- before the fix, the exact reporter-offset test returned `哎`
- focused boundary tests: 2 passed
- realtime/postprocess/file-finalization/docs suites: 141 passed, 1 existing environment skip
- Python compilation and `git diff --check`: passed

Related to #3302. Merging this PR is not reporter confirmation; keep the issue open until the original file-upload and microphone paths are retested.

Signed-off-by: LauraGPT <18321252+LauraGPT@users.noreply.github.com>

### DIFF
--- a/funasr/bin/realtime_ws.py
+++ b/funasr/bin/realtime_ws.py
@@ -1099,7 +1099,8 @@ class RealtimeASRSession:
                 and recent_eligible
                 and recent_observations >= 2
                 and int(recent_start_ms) == int(seg[0])
-                and 0 <= tail_gap_ms <= max(100, decode_chunk_ms * 2)
+                and -max(100, decode_chunk_ms) <= tail_gap_ms
+                and tail_gap_ms <= max(100, decode_chunk_ms * 2)
                 and recent_partial
                 and not recent_hallucinated
             ):
--- a/tests/test_realtime_ws_service.py
+++ b/tests/test_realtime_ws_service.py
@@ -569,6 +569,48 @@ def test_long_segment_recovers_from_single_character_final():
     assert text == partial
 
 
+def test_long_segment_accepts_partial_slightly_past_vad_end():
+    module = load_service_module()
+    partial = (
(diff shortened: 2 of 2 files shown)


LauraGPT:
Revalidated the exact head `980765099dea8d9f34ea2f10889b3ab1c3876173` in a clean detached worktree.

    PYTHONPATH=. python -m pytest -q tests/test_realtime_ws_service.py tests/test_websocket_file_finalization.py
    # 114 passed

This covers the bounded partial-VAD-overrun rule together with offline/online/2-pass file finalization and completion contracts. Its scope remains intentionally narrow: it prevents an observed complete partial from being discarded when it only slightly exceeds the server VAD boundary; it does not claim to resolve unrelated short-segment recognition drift or every hotword-quality report. Ready for maintainer review.

PR #3592, opened 2026-08-31
feat(qwen3-asr): add offline vLLM example

LauraGPT:
## Summary

- add an offline Qwen3-ASR vLLM example based on the native `Qwen3ASRModel.LLM` backend
- normalize arbitrary input audio to mono 16 kHz, use qwen-asr native chunking, and retain exact chunk timestamps in JSON output
- add Chinese and English operational notes, including the distinction from FunASR `AutoModelVLLM` and accuracy-comparison boundaries
- add focused tests for timestamps, result-count validation, lazy optional dependencies, and CLI defaults

Related to #3419. This PR intentionally does not close the issue because the reporter still needs to validate the workflow in their environment.

## Verification

- `9 passed`: the new tests plus existing Qwen3-ASR WebSocket and dependency-check tests
- `py_compile`, CLI `--help`, and `git diff --check` passed
- real NVIDIA H100 80 GB run with `Qwen/Qwen3-ASR-1.7B` on the 480.24 s #3419 attachment produced three monotonic chunks covering 0-480.216 s
- output evidence SHA256: `899d0eacc8db73e22cbe7e7aeb8befd56c8cc775898862aa327dc4d90a7df7aa`

## Accuracy note

The earlier 18.74% prefix-trim CER used the original reference and a standard character normalization. The reporter later supplied an updated reference containing the first 17.69 seconds and their exact scorer (SHA256 `66ec56383c505418c78c5e7fe567d33564fcce7faccb0f508959a57f76d4ae50`), which additionally normalizes numbers and ignores selected one-character fillers and substitutions. Under that exact custom scoring contract, the PR's native vLLM output scores 6.45% CER; the separately tested Transformers pipeline scores 6.80%, and the 20-second VAD/vLLM path scores 6.63%. These results reproduce the reporter's stated 8-9% range but are not directly comparable with the earlier standard-normalization result or an official benchmark.

## Attribution

The architecture was cross-checked against the MIT-licensed `LanceLRQ/qwen3-asr-service` project, which is acknowledged in the accompanying notes.

### DIFF
--- a/examples/industrial_data_pretraining/qwen3_asr/transcribe_vllm_offline.py
+++ b/examples/industrial_data_pretraining/qwen3_asr/transcribe_vllm_offline.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+"""Transcribe long audio with Qwen3-ASR's native vLLM backend."""
+
+import argparse
+import json
+import subprocess
+import tempfile
+from pathlib import Path
+
+
+SAMPLE_RATE = 16000
+
+
+def _result_value(result, key, default=None):
+    if isinstance(result, dict):
+        return result.get(key, default)
+    return getattr(result, key, default)
+
+
+def transcribe_chunks(model, chunks, sample_rate, language):
+    """Transcribe qwen-asr chunks and restore offsets on the source timeline."""
+    segments = []
+    for audio, offset_seconds in chunks:
+        results = model.transcribe(audio=(audio, sample_rate), language=language)
+        if len(results) != 1:
+            raise RuntimeError(
+                f"Qwen3-ASR must return one result per chunk, received {len(results)}"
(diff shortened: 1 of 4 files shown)


LauraGPT:
Revalidated exact head `1bd0b24fa853becd0b8c186730c1cf61535b59ee` in an isolated worktree:

    PYTHONPATH=. python -m pytest -q tests/test_qwen3_asr_vllm_offline_example.py
    # 3 passed

`git diff --check origin/main...HEAD` is also clean. The contract covers the native offline vLLM request construction, audio normalization/chunk boundaries, and JSON segment schema. The example is ready for review; its benchmark note remains scoped to the linked audio and scorer rather than claiming a universal CER.

PR #3674, opened 2026-09-08
fix: allow numpy 2.x by replacing removed type aliases

linhongyu510:
## Summary

Related to #3673. Keep that issue open for the affected installation/runtime re-test; this source change is not yet in the published PyPI 1.4.14 package.

Retains @linhongyu510's original NumPy alias cleanup and dependency-cap removal, with maintainer follow-up covering the real scientific-stack paths before expanding the supported range:

- CMVN uses `np.float64`; emotion2vec's non-overlapping mask allocator uses `np.int64`.
- Runtime tests now invoke the actual mask function, frontend initialization/forward pass, and fresh-process module imports. They do not duplicate an isolated `np.fromiter` expression or rely only on cached top-level imports.
- Exercises real SciPy/sklearn spectral clustering and the UMAP/HDBSCAN wrapper using synthetic embeddings, with label-permutation-invariant partition assertions.
- Fixes six existing EEND mel-filter calls to use librosa's keyword-only arguments. This was reproduced on the old main branch as well as the original PR: it is a pre-existing modern-librosa incompatibility, not a regression caused by NumPy 2.
- Adds a CPU CI matrix with the exact NumPy versions below, resolved dependency reports, `pip check`, source commit, and JUnit artifacts.

## Verification

Local exact-head verification: `9d11d2c4adef9d723e192b224194bd982488bde7`, incorporating main `01e822be1584ecb51e306c12f53b2dd47dcb96db`.

Both are separate clean Linux/Python 3.12 CPU environments with Torch/TorchAudio 2.10.0. Optional `pytorch-wpe` is installed to exercise the multichannel frontend module; no model weights or GPU inference are involved.

| NumPy | Resolved SciPy | Resolved librosa | Compatibility suite | Expanded local KWS/VAD regression |
| --- | --- | --- | --- | --- |
| 1.26.4 | 1.17.1 | 0.11.0 | 23 passed, 0 skipped | 77 passed + 184 subtests passed, 0 skipped |
| 2.4.0 | 1.18.1 | 1.0.0 | 23 passed, 0 skipped | 77 passed + 184 subtests passed, 0 skipped |

`pip check` passes in both environments. A resolver countercheck against old main plus `numpy>=2.4.0` produces the reported `ResolutionImpossible`; the candidate resolves successfully. With the new runtime tests retained, old main fails the real CMVN/masking alias paths (5 failures, 1 import check passing after supplying optional frontend dependencies). The seven EEND checks fail before the keyword-argument fix and pass afterward.

The historical justification for the broad NumPy cap was not independently established as being *only* an alias issue. This PR's evidence is the actual compatibility and resolver matrix above, not a blanket ABI claim. The initial local installation hit a server package-source DNS failure; that failed attempt was retained and installation was completed using an explicit public index.

## Boundaries

- These tests do not establish GPU/model accuracy, multichannel WPE numerical quality, the >=2048-embedding dispatcher, or every historical Python/scientific-library combination.
- Three existing warnings remain visible: Torch STFT's deprecated real-pair return, the EEND frontend's read-only NumPy view passed to Torch, and sklearn HDBSCAN's future `copy` default. No warnings, tests, or tolerances were suppressed to obtain a pass.
- GitHub CI must be checked on the exact head before merging. Local results are not a substitute for that gate.


LauraGPT:
Thanks for addressing the removed NumPy aliases. I reviewed the current head (`a8edde15f72b1ec7cd8635deff7d2c6811e10592`). The float64 replacement looks appropriate; the remaining merge concern is the evidence for removing the NumPy upper bound, not a demonstrated universal NumPy 2 failure.

Before expanding the supported dependency range, we need:
1. A test that calls the real `compute_mask_indices(no_overlap=True)` path with padding/spacing checks. The current `np.fromiter` test duplicates an expression, so a regression in the production function would not fail it.
2. A fresh-process import check for the relevant modules. Top-level `import funasr` records optional submodule exceptions by default and can already be cached by the CMVN test; the NumPy attribute assertions do not establish a successful frontend import.
3. CI coverage on compatible Python for NumPy 1.26 and 2.x, exercising the frontend, masking and NumPy/Torch array bridge with resolved dependency versions and `pip check`. This head currently has no check runs, and existing KWS CI still pins `numpy<2`.

The alternative is to land the alias cleanup first while retaining the dependency cap until the compatibility matrix is verified. No need for model weights to strengthen these specific regression tests. The linked issue should remain open until the affected installation/runtime path is verified.

PR #3703, opened 2026-09-11
fix(nano): correct punctuation timestamps across VAD merges

### LINKED ISSUE #3702: Fun-ASR-Nano模型打开merge_vad参数后标点符号的时间戳不对
使用Fun-ASR-Nano模型打开merge_vad参数后，标点符号的时间戳计算不对
本句话的标点符号的时间戳计算到下一句话开始的时间上了

Bruce-Yii:
Closes #3702.

## Problem

With Fun-ASR-Nano and `merge_vad` enabled, a sentence-final punctuation mark's timestamp lands on the next sentence's start time.

## Root cause

`merge_vad` intentionally merges neighboring VAD regions (including the silence gaps between them) into one ASR chunk, so a single decode window can span several sentences. Nano emits punctuation natively and its per-token times come from CTC forced alignment (`forced_align`), which must place **every** target token — including punctuation tokens with zero acoustic realization — on at least one frame. The alignment puts such a token at the next sentence's onset frame. That span is then propagated 1:1 to the user-visible `timestamp` list (frame→ms scaling plus chunk-offset restoration), with no correction anywhere on the Nano path (the separate punctuation model is skipped by design for Nano). The same shared `forced_align` feeds the offline path, the vLLM path, and the vLLM pipeline path.

## Fix and invariant

Punctuation carries no acoustic extent. `anchor_punctuation_timestamps` (new, in `funasr/models/fun_asr_nano/tools/utils.py`) pins a punctuation span that is followed by more speech to the preceding spoken token's end (zero-width). It uses one explicit spoken/punctuation/special classifier so `<sil>`-style special tokens neither trigger anchoring nor serve as anchors. Leading/trailing punctuation and special tokens are left untouched, preserving single-segment behavior. Wired into all three Nano timestamp surfaces: `FunASRNano.inference_llm`, `FunASRNanoVLLM._compute_timestamps`, and `FunASRNanoVLLMPipeline._compute_all_timestamps` — the pipeline applies it per independently aligned VAD segment, so speech in a later segment can never rewrite a prior segment's trailing punctuation. `ctc_timestamps` is intentionally untouched.

## Validation performed

- New `tests/test_fun_asr_nano_punctuation_timestamps.py`: 8/8 pass, including a stubbed-engine behavior test of the pipeline path (no weights/GPU) proving intra-segment anchoring plus segment-boundary preservation.
- Neighboring offline tests (`test_timestamp_tools.py`, `test_fun_asr_nano_missing_ctc_weights.py`): 15/15 pass (23/23 total with the new file).
- `git diff --check` clean; added lines follow repo conventions (black hunks verified to avoid them).
- Deterministic forced-alignment reproduction with synthetic peaky emissions through the repo's exact `forced_align`: merged-window punctuation lands at the silence end (next-sentence onset) while the single-sentence control keeps it at the window end; after anchoring it sits at the sentence end.
- Full test-suite validation was not run, and real-weight Nano inference was not run; no such claims are made.

### DIFF
--- a/funasr/models/fun_asr_nano/tools/utils.py
+++ b/funasr/models/fun_asr_nano/tools/utils.py
@@ -83,6 +83,11 @@ def forced_align(log_probs: torch.Tensor, targets: torch.Tensor, blank: int = 0)
 
 _PUNCTUATION_TOKEN_RE = re.compile(r"^[^\w\s]+$")
 _SPECIAL_TOKEN_RE = re.compile(r"^<[^>]*>$")
+# tiktoken decodes a partial UTF-8 byte sequence (one id of a multi-id
+# character, e.g. SenseVoice ids [10958, 245] for 郗) as U+FFFD. All three
+# Nano timestamp paths decode each token id independently, so such fragments
+# reach classification as "�". They are spoken content, never punctuation.
+_UNDECODABLE_FRAGMENT_MARK = "�"
 
 
 def _classify_timestamp_token(token):
@@ -90,13 +95,18 @@ def _classify_timestamp_token(token):
 
     Special tokens (``<sil>``-shaped) are their own class: they are neither
     spoken content nor punctuation, so they neither trigger anchoring nor
-    serve as anchors. Non-string tokens (integer ids, if called pre-decode)
-    fall back to spoken to preserve pass-through behavior.
+    serve as anchors. Undecodable byte fragments (containing U+FFFD) are
+    spoken: they are slices of a multi-id character with real acoustic
+    extent, and collapsing their spans would delete spoken timing (issue
+    #3703). Non-string tokens (integer ids, if called pre-decode) fall back
+    to spoken to preserve pass-through behavior.
     """
     if not isinstance(token, str) or not token:
         return "spoken"
     if _SPECIAL_TOKEN_RE.match(token):
         return "special"
+    if _UNDECODABLE_FRAGMENT_MARK in token:
+        return "spoken"
     if _PUNCTUATION_TOKEN_RE.match(token):
         return "punctuation"
     return "spoken"
--- a/funasr/models/fun_asr_nano/tools/utils.py
+++ b/funasr/models/fun_asr_nano/tools/utils.py
@@ -1,4 +1,5 @@
 from itertools import groupby
+import re
 
 import librosa
 import soundfile as sf
@@ -78,3 +79,66 @@ def forced_align(log_probs: torch.Tensor, targets: torch.Tensor, blank: int = 0)
     except Exception:
         pass
     return items
+
+
+_PUNCTUATION_TOKEN_RE = re.compile(r"^[^\w\s]+$")
+_SPECIAL_TOKEN_RE = re.compile(r"^<[^>]*>$")
+
+
+def _classify_timestamp_token(token):
+    """Classify one decoded timestamp token into spoken/punctuation/special.
+
+    Special tokens (``<sil>``-shaped) are their own class: they are neither
+    spoken content nor punctuation, so they neither trigger anchoring nor
+    serve as anchors. Non-string tokens (integer ids, if called pre-decode)
+    fall back to spoken to preserve pass-through behavior.
+    """
+    if not isinstance(token, str) or not token:
+        return "spoken"
+    if _SPECIAL_TOKEN_RE.match(token):
+        return "special"
+    if _PUNCTUATION_TOKEN_RE.match(token):
+        return "punctuation"
+    return "spoken"
+
+
+def anchor_punctuation_timestamps(timestamps):
+    """Anchor punctuation-only tokens at the preceding spoken token's end.
+
+    CTC forced alignment must place every target token — including punctuation
+    tokens with no acoustic realization — on at least one frame. Inside a
+    merged multi-sentence window (see FunASR issue #3702) such a token lands
+    at the next sentence's onset frame, so the current sentence's punctuation
+    timestamp reads as the next sentence start. Punctuation carries no acoustic
+    extent, so a punctuation span that is followed by more speech is pinned
+    (zero-width) at the end of the speech it terminates.
+
+    Leading punctuation (no predecessor) and trailing punctuation (no
+    successor, e.g. single-segment output) are left untouched, as are special
+    tokens such as ``<sil>``. This preserves existing single-segment behavior.
+
+    Args:
+        timestamps: List of ``{"token": str, "start_time": float,
+            "end_time": float}`` dicts in time order. Mutated in place.
+
+    Returns:
+        The same list, for convenient chaining.
+    """
+    last_spoken_end = None
+    total = len(timestamps)
+    for index, ts in enumerate(timestamps):
+        kind = _classify_timestamp_token(ts.get("token", ""))
+        if kind != "punctuation":
+            if kind == "spoken":
+                end = ts.get("end_time")
+                if isinstance(end, (int, float)):
+                    last_spoken_end = end
+            continue
+        following_spoken = any(
+            _classify_timestamp_token(later.get("token", "")) == "spoken"
+            for later in timestamps[index + 1 : total]
+        )
+        if last_spoken_end is not None and following_spoken:
+            ts["start_time"] = last_spoken_end
+            ts["end_time"] = last_spoken_end
+    return timestamps
--- a/funasr/models/fun_asr_nano/inference_vllm.py
+++ b/funasr/models/fun_asr_nano/inference_vllm.py
@@ -694,7 +694,10 @@ class FunASRNanoVLLM:
     @torch.no_grad()
     def _compute_timestamps(self, encoder_out, encoder_out_lens, text):
         """CTC forced alignment for character-level timestamps."""
-        from funasr.models.fun_asr_nano.tools.utils import forced_align
+        from funasr.models.fun_asr_nano.tools.utils import (
+            anchor_punctuation_timestamps,
+            forced_align,
+        )
 
         decoder_out, decoder_out_lens = self.ctc_decoder(encoder_out, encoder_out_lens)
         ctc_logits = self.ctc.log_softmax(decoder_out)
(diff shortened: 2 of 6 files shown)


funasr/models/fun_asr_nano/tools/utils.py
```diff
@@ -77,4 +78,67 @@ def forced_align(log_probs: torch.Tensor, targets: torch.Tensor, blank: int = 0)
...
+    Special tokens (``<sil>``-shaped) are their own class: they are neither
+    spoken content nor punctuation, so they neither trigger anchoring nor
+    serve as anchors. Non-string tokens (integer ids, if called pre-decode)
+    fall back to spoken to preserve pass-through behavior.
+    """
+    if not isinstance(token, str) or not token:
+        return "spoken"
+    if _SPECIAL_TOKEN_RE.match(token):
+        return "special"
+    if _PUNCTUATION_TOKEN_RE.match(token):
```
LauraGPT, line 110: [P2] Do not classify undecodable spoken byte fragments as punctuation. Nano's real CTC tokenizer is byte-based, and all three new call sites classify decode([token_id]) independently. For the ordinary surname in `我叫郗明。你好。`, the official vocabulary encodes 郗 as IDs [10958, 245]: decoding them together gives 郗, but each separately gives U+FFFD (`�`). This regex classifies both as punctuation. With actual forced_align on deterministic emissions, their valid spans [0.60, 0.72] and [0.84, 0.96] seconds are both changed to [0.48, 0.48], the preceding 叫's end. The actual vLLM method reproduces this, and the pipeline reproduces it after its +1s VAD offset (both become [1.48, 1.48]). Thus this patch deletes spoken timing information for byte-split characters, not merely punctuation. Please base classification on valid decoded text/token provenance (at minimum, undecodable fragments must not be treated as punctuation), and add a regression using the official tokenizer plus multi-ID spoken characters across the timestamp methods. The current tests use a one-ID-to-one-character stub and cannot catch this. Full-sequence decoding works; the pre-existing per-ID display replacement is not itself introduced by this PR, but collapsing its valid acoustic spans is.

LauraGPT (changes requested):
I ran the three reported test files on exact head 4552e01d695e7309aa29dde95a704294325c9912: 23 passed. I then checked the token-classification assumption against Nano's actual SenseVoiceTokenizer using the official multilingual.tiktoken at fixed HF revision 272c57b82523ada6fd87095e955f8e29100979ab (SHA256 747979631e813193436aabcff7c1c235d37de8097b71c563ec8b63b7a515c718, verified against the public artifact). This uncovered a spoken-token timestamp regression, detailed inline.

The additional probe uses real tokenization and the repository's actual forced_align with deterministic synthetic log probabilities. It has 3 failing preservation assertions (helper, actual vLLM timestamp method, actual pipeline timestamp method) and 2 passing controls (tokenizer/alignment roundtrip and genuine sentence punctuation anchoring). The neural encoder/decoder/frontend are synthetic stand-ins; no weights, GPU, acoustic accuracy, full offline inference or vLLM-engine acceptance is claimed. Please add a real-tokenizer byte-fragment regression and address classification before merging. Issue #3702 should remain open until the revised behavior is verified.

funasr/models/fun_asr_nano/tools/utils.py
```diff
@@ -77,4 +78,77 @@ def forced_align(log_probs: torch.Tensor, targets: torch.Tensor, blank: int = 0)
...
+    #3703). Non-string tokens (integer ids, if called pre-decode) fall back
+    to spoken to preserve pass-through behavior.
+    """
+    if not isinstance(token, str) or not token:
+        return "spoken"
+    if _SPECIAL_TOKEN_RE.match(token):
+        return "special"
+    if _UNDECODABLE_FRAGMENT_MARK in token:
+        return "spoken"
+    if _PUNCTUATION_TOKEN_RE.match(token):
```
LauraGPT, line 110: [P2] Preserve semantic symbols instead of classifying every non-word character as punctuation. The U+FFFD guard fixes undecodable fragments, but this regex still accepts valid Unicode symbols. With the same verified official SenseVoice vocabulary, `我用C++。你好。` encodes `++` as ID 24754 (Unicode Sm), and `价格是$10。你好。` encodes `$` as ID 3 (Sc). Both decode without replacement characters. Using real forced_align on deterministic peaked emissions and the actual FunASRNanoVLLM._compute_timestamps, each symbol's [0.84, 0.96] seconds becomes [0.72, 0.72]. These symbols can represent spoken semantic content, not sentence punctuation, so this discards an existing alignment span outside the intended correction. Please use an explicit supported sentence-punctuation policy (or another conservative semantic classifier) rather than the complement of word/whitespace characters, and add real-tokenizer C++/currency controls alongside the genuine punctuation and byte-fragment tests. These are synthetic-emission reproductions, not claims about a particular recording or acoustic accuracy.

LauraGPT (changes requested):
Rechecked exact head 510b81ee7149b21740b2353bb749af69f77e3289. The previous U+FFFD byte-fragment finding is fixed: all four relevant repository test files plus the unchanged five-test official-vocabulary review probe passed (32 total). The original two spoken-fragment spans are preserved in the helper, actual vLLM timestamp method and pipeline method, including the VAD offset. I am resolving that original thread.

A further check of the same classifier found a distinct remaining problem with valid mathematical/currency symbols, detailed inline. Two official-tokenizer inputs reproduce span deletion in the actual vLLM timestamp method. Please restrict anchoring to intentionally supported sentence punctuation and add symbol-preservation coverage before merge.

All checks here are CPU-side tests with real tokenizer/forced_align and synthetic emissions; neural components are stand-ins. No real-weight Nano inference, vLLM engine, acoustic accuracy or full-suite acceptance is claimed. Issue #3702 remains open.

funasr/models/fun_asr_nano/tools/utils.py
LauraGPT: Rechecked exact head `48ef4b2f1b4c81903602c59e3283fcb06c1fd9a3`. The explicit sentence-punctuation set fixes this reported semantic-symbol regression: `++` (official ID 24754) and `$` (ID 3) retain their original nonzero spans in the helper, actual vLLM timestamp method, and actual pipeline timestamp method, including the +1s VAD offset. Genuine sentence punctuation still anchors, and the byte-fragment preservation controls remain green.

Validation: **44 passed** on CPU: 33 tests across the five relevant repository files, the unchanged five-test official-vocabulary byte-fragment review probe, and six official-vocabulary symbol probes. The cached official vocabulary was SHA-256 checked (`747979631e813193436aabcff7c1c235d37de8097b71c563ec8b63b7a515c718`), and all seven changed files matched the exact head Git blobs before testing and their hashes afterward. This uses real tokenization/forced_align with synthetic emissions and stand-in neural components; it is not real-weight inference, vLLM-engine, GPU, acoustic-accuracy, full-offline-path or full-suite acceptance.

I am resolving this specific inline finding. Separately, GitHub currently reports `mergeable=false / mergeable_state=dirty`: please synchronize the branch with the current target and resolve conflicts without dropping these regressions. That resulting integration tree still needs checks; these results do not approve an untested merge tree or close #3702.

Prepared with Codex assistance.

Bruce-Yii:
Synced with current `main` (`02f8b43f`) locally and resolved the one real conflict: the `model.py` import block now keeps both upstream `configure_llm_forward` and this PR's `anchor_punctuation_timestamps`/`forced_align` (orthogonal additions, no semantic change). New head `3b0122c7`; the 7 PR files are otherwise byte-identical to the reviewed head. Revalidation on the integration tree: 33/33 focused repo tests, byte-fragment spans preserved (helper/vLLM/pipeline, official tokenizer), `C++`/`$` symbol spans preserved with genuine `。` still anchoring, adjacent `train_opts` 7 passed + 2 CUDA-only skipped, `git diff --check` clean. No scope change; awaiting review/CI.

LauraGPT (commented):
Rechecked exact head `3b0122c733d5ceac69f72ef0d5b6d68d346cf030` against current `main@02f8b43fc9222e3d14d41565acbebbb23bf3b56c` after the conflict sync.

The reviewed PR files remain byte-identical to `48ef4b2f` except for `model.py`, where the merge correctly preserves both upstream `configure_llm_forward` and this PR's `anchor_punctuation_timestamps` / `forced_align` imports and call. I found no new code issue in that resolution.

Independent verification on the exact head passed:

- focused timestamp surface: 33/33;
- adjacent `test_fun_asr_nano_train_opts.py`: 7 passed, 2 CUDA-only skipped;
- compileall and `git diff --check`.

The remaining merge gate is commit provenance. GitHub reports the new merge head as unsigned, it has no `Signed-off-by` trailer, and its subject says “local only, do not push.” Please rewrite that merge commit (preserving the tested tree) with a normal subject, a verified signature, and DCO sign-off, then update the PR. Please do not add an empty follow-up commit. I will recheck the new exact head rather than asking you to repeat the already-green test scope without a code change.

Bruce-Yii:
Provenance-only rewrite complete as requested: new head `f7f5c227` replaces `3b0122c73` with a normal subject, DCO sign-off, and a verified SSH signature. Tree (`8d752845`) and ordered parents (`48ef4b2`, `02f8b43f`) are unchanged — no code or test changes. No further action from my side unless review finds otherwise.

LauraGPT (approved):
Rechecked exact head `f7f5c227999668dd325cf85923f07a9ef6786bb3` following the requested provenance-only rewrite.

GitHub's Git objects independently confirm that the complete tree is still `8d752845d06d67f15253aba6b0205194c8b95603`, identical to the previously reviewed integration head `3b0122c733d5ceac69f72ef0d5b6d68d346cf030`. The ordered parents are also unchanged: `48ef4b2f1b4c81903602c59e3283fcb06c1fd9a3` and current main `02f8b43fc9222e3d14d41565acbebbb23bf3b56c`. No source or test changes were introduced by this rewrite.

The new merge commit has a normal subject, a GitHub-verified SSH signature, and the author's DCO sign-off. This satisfies the specific merge-head provenance request in my previous review; it does not claim that the older commits were retroactively signed. Both earlier byte-fragment and semantic-symbol review threads are resolved, and their fixes are preserved in the identical tree. I have no remaining code-review finding and am approving that reviewed change.

The prior exact-tree CPU review results remain the recorded evidence: 33 focused tests plus 7 adjacent training-option tests passed, with 2 CUDA-only tests skipped. I did not rerun those tests for this metadata-only rewrite, and this is not a real-weight inference, GPU, vLLM-engine, acoustic-accuracy, or full-suite acceptance claim.

CI remains separate: current-head workflow `35687719856` is `action_required`, with no check runs. This approval is not a claim of green CI or authorization to bypass workflow approval, merge protections, or release gates. Keep #3702 open pending publicly available delivery and verification of the reporter's workflow.

Prepared with Codex assistance.

PR #3709, opened 2026-09-15
Fix invalid escape sequences in text normalization and models

MBK-fr:
### Problem

Three first-party modules use invalid escape sequences in non-raw string literals:

- `fun_text_processing/text_normalization/normalize.py` - `\.`, `\w`
- `funasr/models/fsmn_kws/encoder.py` - `\[`
- `funasr/models/fun_asr_nano/tools/format5res.py` - `\.`

On Python 3.12+ each raises `SyntaxWarning: invalid escape sequence` on import.

### Why it matters for FunASR

FunASR declares `requires-python >=3.7.0` and runs the 3.10-3.12 CI matrix, so the warnings surface on the 3.12 job.

### Fix

Escape the backslashes; every string value (and the regexes built from them) is byte-for-byte identical (ruff W605-style). (An identical copy of `format5res.py` under `examples/` was left untouched to keep this scoped to the package source.)

### Verification

In-memory `compile()` under `-W error` is clean on all three files after (raised before); each module's AST is unchanged.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014HBQNzAf5C2E3MiJC48HQw


LauraGPT (commented):
I independently checked the exact three changed files at `7cae84872c088166b544d789de29632f2a7911dc` against base `486b4b7ceb27b72db24a84c6e1be7cf6c5be6609`, using Python 3.12.3.

With warnings treated as errors, in-memory compilation reproduces the invalid-escape failure in all three base files, while all three candidate files compile cleanly. Comparing complete module ASTs with location attributes excluded gives identical trees for each pair, including the interpolated normalization pattern and string constants. No behavioral discrepancy was found within this scope.

This validates the submitted package-source cleanup, not all repository copies or all supported Python versions. As noted in the PR description, the separate example copy of `format5res.py` is unchanged. I did not import these modules, load model weights, or run inference/full runtime tests. This is a bounded verification update, not a merge or full-runtime approval.

Issue #3253, opened 2026-07-19
生成字幕功能完全不断句，只会直接返回全文


dearvcc:
## 🐛 Bug

根据文档，下面的命令可以生成字幕文件：
funasr tst.mp4 --output-format srt -o subs --lang zh

但结果是巨长的一整句，如图所示。

已经升到最新版：
funasr --version
funasr-1.3.17

<img width="3813" height="1245" alt="Image" src="https://github.com/user-attachments/assets/a8576250-b905-4012-bf82-843326a96f35" />

LauraGPT:
Thanks for the clear report. I reproduced the CLI-side cause: the top-level `funasr ... --output-format srt` path was not requesting sentence-level timestamps, and the default SenseVoice CLI config did not load a punctuation model for subtitle generation, so the SRT formatter could only fall back to one full-text cue.

I opened a fix in #3254. It makes SRT/TSV generation request sentence timestamps, loads `ct-punc` for subtitle output when needed, and keeps a valid timestamp/audio-bound fallback. The focused CLI/subtitle tests pass locally.

Before this PR is merged/released, a workaround is to use the subtitle example path that already requests sentence timestamps, for example:

```bash
python examples/subtitle/generate_subtitle.py tst.mp4 -o subs/tst.srt --lang zh
```

PR #3726, opened 2026-09-21
fix(cli): warn when jieba is unavailable for readable SRT splitting

Linxiushen:
## Problem

`funasr audio.wav -f srt` (default `--subtitle-segment-mode readable`) scores candidate split points with `_subtitle_break_weights()` in `funasr/cli.py`, which uses jieba for lexical boundaries. When `import jieba` fails, the function catches the error and silently continues with punctuation-only weights:

```python
except (ImportError, RuntimeError, ValueError):
    pass
```

For continuous Chinese speech without punctuation, every candidate split then carries the same penalty and the balanced-range DP degenerates to splitting by character count, so cues break in the middle of words. That is precisely the behaviour `test_merge_subtitle_segments_avoids_single_character_phrase_breaks` guards against, and in an environment without jieba that test genuinely fails. Nothing is logged, so a user seeing broken cues cannot tell an environment problem from a model problem.

Reachability, stated plainly: jieba is in `install_requires` and in the PyPI wheel's `requires_dist`, so a standard `pip install funasr` is not affected. The fallback is reached with `pip install --no-deps`, slimmed Docker images, or a broken/conflicting environment. In those cases the degradation is fully silent.

## Change

Log one warning per process when the fallback is taken, using the module-level `logging.warning(...)` style already used elsewhere in the package (`funasr/utils/timestamp_tools.py`, `funasr/download/download_model_from_hub.py`). The `import logging` inside the `try` block moves to the top of the file. Output is unchanged.

The value is diagnosability: subtitle segmentation quality is a recurring issue topic here (#3253, #3406, #3539, #3644), and this line lets a maintainer rule out a missing jieba immediately.

## Verification

- New test `test_subtitle_break_weights_warns_once_without_jieba` blocks `jieba` via `sys.modules`, calls the function twice, and asserts identical weights and exactly one warning. It fails on main (`assert 0 == 1`) and passes with the change.
- Output differential: for 14 texts (unpunctuated Chinese, punctuated Chinese, English, mixed, Korean, Japanese, single character, empty, punctuation-only) plus two multi-segment cases, the serialized outputs of `_subtitle_token_spans`, `_subtitle_break_weights`, `merge_subtitle_segments` and `_balanced_subtitle_ranges` are byte-identical before and after, both with and without jieba installed.
- End to end: `-f srt -o ./subs` writes an identical `.srt`; `-f srt` to stdout is byte-identical, the warning goes to stderr only, so `funasr a.wav -f srt > out.srt` pipelines are unaffected.
- `tests/test_cli.py`: 24 passed, 4 failed both before and after; the 4 failures are pre-existing in my environment (missing `kaldiio`/`librosa`) and unrelated.

---
This fix was developed with AI assistance (Claude); the change and the test were reviewed and verified locally before submission.


LauraGPT (commented):
Reviewed exact head `2b69a795ee9f57f009a558470179847421accade` against base `02f8b43fc9222e3d14d41565acbebbb23bf3b56c`. I found no blocking code issue in this diagnostic-only change.

Independent checks with Python 3.12.3 and the actual source files:

- Current head: all 28 tests in `tests/test_cli.py` passed.
- Base implementation with the existing tests: 27 passed.
- The new regression test against the base implementation failed at `warning.call_count == 1` with `assert 0 == 1`. The unchanged-weight assertion passed first.

The four dependency-related failures reported in the PR description did not occur in this environment. These checks use the tests' mocked ASR models; they are not audio-recognition or full-suite acceptance. I did not independently repeat the separate 14-text differential experiment described in the PR.

The current commit is unsigned and has no `Signed-off-by` trailer. Please add a verified commit signature and DCO sign-off before the merge check. I am leaving this review as `COMMENTED` and am not merging the unsigned commit.

Linxiushen:
Thanks for the careful review, @LauraGPT — and for re-running it rather than taking my word for it. Your numbers match what I saw: 28 passed on the head, 27 on the base, and the new test failing on the base at `warning.call_count == 1` with `assert 0 == 1`.

Two clarifications on your notes:

- The four dependency-related failures are environment-only, as you found. They come from `kaldiio` and `librosa` being absent in my local venv, so lazy imports fail during those tests. They are unrelated to this change and I would not expect them anywhere those packages are installed. Sorry for the noise in the description — I listed them to be transparent about what my run looked like, not to suggest they were caused by the patch.
- The 14-text differential is the part that matters most for the "no behaviour change" claim, so it is fair that you did not take it on trust. For anyone repeating it: it dumps `_subtitle_token_spans`, `_subtitle_break_weights`, `merge_subtitle_segments` and `_balanced_subtitle_ranges` for each text as sorted JSON and compares hashes with and without the patch, both with and without `jieba` importable. The point is that the only difference in the output is the single stderr line.

On the signature and sign-off: those are the account owner's own certification to make, so I am not going to add a `Signed-off-by` trailer on their behalf. I have passed your request on and the signed, signed-off commit will come from the account owner rather than from me. Understood that the PR is not mergeable until then.

